PR #2, opened 2016-05-12
object api tests

diasdavid:
**WIP**.. this is going to be big

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -3,6 +3,9 @@
 'use strict'
 
 const expect = require('chai').expect
+const DAGNode = require('ipfs-merkle-dag').DAGNode
+const bs58 = require('bs58')
+const jsonToYaml = require('json2yaml')
 
 module.exports = (common) => {
   let ipfs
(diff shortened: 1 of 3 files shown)


src/object.js
```diff
@@ -25,7 +26,31 @@ module.exports = (common) => {
...
+        expect(err).to.not.exist
+        console.log('->', node)
+        // expect(res).to.have.a.property('Hash', testObjectHash)
+        // expect(res.Links).to.be.empty
+        done()
+      })
+    })
+
+    it.skip('object.put of json encoded buffer', (done) => {
+      ipfs.object.put(testObject, 'json', (err, res) => {
```
dignifiedquire: also need tests for 

```js
```

dignifiedquire:
Also need tests for

```js
ipfs.object.put(new Buffer('hello'), {enc: 'json'})
ipfs.object.put(new Buffer('hello'), {enc: 'yaml'})
```

diasdavid:
> ipfs.object.put(new Buffer('hello'), {enc: 'json'})

Not sure how this would work, if we say that the Buffer has json or yaml encoding, shouldn't it be complete? Passing it through the HTTP-API makes it fail, I've made a test for it:

https://github.com/ipfs/interface-ipfs-core/pull/2/files#diff-eaac2b5121380ed7f166ff7fe9e32c5bR108

The test that I have and that works as expect is: 

https://github.com/ipfs/interface-ipfs-core/pull/2/files#diff-eaac2b5121380ed7f166ff7fe9e32c5bR45

src/object.js
```diff
@@ -20,49 +22,264 @@ module.exports = (common) => {
...
+        done()
+      })
+    })
+
+    // TODO checking if this is what we really want
+    // https://github.com/ipfs/interface-ipfs-core/pull/2#issuecomment-218980224
+    it.skip('object.put of buffer and enc json', (done) => {
+      const buf = new Buffer('Some data')
+
+      ipfs.object.put(buf, { enc: 'json' }, (err, node) => {
```
dignifiedquire: Sorry, I meant sth like this:

```js
const buf = new Buffer(JSON.stringify({Data: 'hello world', Links: ['some link']}))
```

dignifiedquire: and also the same with a yaml version

diasdavid: Ok, the first one is already there, just missing the yaml one then

src/object.js
```diff
@@ -20,233 +23,599 @@ module.exports = (common) => {
...
+          expect(expected).to.deep.equal(stats)
+          done()
+        })
+      })
+    })
+
+    it('object.stat with multihash base58 encoded toString', (done) => {
+      const testObj = {
+        Data: new Buffer('get test object'),
+        Links: []
```
dignifiedquire: Need a test with links, to ensure the returned stats are correct

diasdavid: good point! thank you :)

diasdavid: added

diasdavid:
tests complete and also js-ipfs-api is compatible with all this tests now :)

diasdavid:
Merging as this code got several reviews by @dignifiedquire :)

src/object.js
```diff
@@ -20,233 +23,700 @@ module.exports = (common) => {
...
+        expect(obj.Links).to.deep.equal(nodeJSON.Links)
+        expect(nodeJSON.Hash).to.equal('QmPb5f92FxKPYdT3QNBd1GKiL4tZUXUrzF4Hkpdr3Gf1gK')
         done()
       })
     })
 
-    it('object.data', (done) => {
-      ipfs.object.data(testObjectHash, (err, res) => {
+    // TODO verify that yaml encoded buffers still work in go-ipfs
+    it.skip('object.put of yaml encoded buffer', (done) => {
```
dignifiedquire: @diasdavid yaml was never available, json and protobuf are the supported encodings according to the help section
```
   Data should be in the format specified by the --inputenc flag.
    --inputenc may be one of the following:
    	* "protobuf"
    	* "json" (default)
```

PR #5, opened 2016-05-13
fix: before and after clauses need to be nested in a describe block

dignifiedquire:
Also added some more blocks to see better grouping of commands


diasdavid:
LGTM, thank you :)

PR #7, opened 2016-05-13
feat(object): object.put protobuf encoding test

### LINKED ISSUE #3: Make a test with encoded protobuf
Following: https://github.com/ipfs/interface-ipfs-core/pull/2#discussion_r63187733

dignifiedquire:
Closes #3

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,7 @@
   "dependencies": {
     "bs58": "^3.0.0",
     "chai": "^3.5.0",
-    "ipfs-merkle-dag": "^0.5.1",
-    "json2yaml": "^1.1.0"
+    "ipfs-merkle-dag": "^0.5.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


diasdavid:
go-ipfs doesn't like the protobuf

```
Uncaught AssertionError: expected [Error: invalid character 'S' looking for beginning of value] to not exist
```

I wonder is because it gets stringified to be put on the wire. @whyrusleeping thoughts?

diasdavid:
Note that we have tests to match the protobufs generated in JS to the ones in Go, byte by byte. Sharing this so that you don't spend time trying to see if the protobuf is correctly formated.

dignifiedquire:
This PR is correct, the issue was in js-ipfs-api see this PR for a fix: https://github.com/ipfs/js-ipfs-api/pull/272

PR #8, opened 2016-05-14
WIP Files

nginnever:
Beginning the files documentation. I haven't added the tests yet but plan to do so after finishing the HTTP-API for JS-IPFS. There are still some things that need to be done to reach complete feature parity with go-ipfs but this is a good place to put the documentation of where js-ipfs is in comparison. 

Starting the gzip compression and tar archiving for jsipfs get command as mentioned in the readme update here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,85 @@ test.all(common)
 
 A valid (read: that follows this interface) IPFS core implementation, must expose the following API.
 
+## Group Files
+
+### `Add`
+
+> Store a Unixfs file or directory of files .
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.add(path, [options, callback])
+
+`path` The path to a file to be added to IPFS. Can be of type:
+
+- Callback, If no first argument is supplied a callback will return a duplex stream.
+- Path, String format path to a file to be added to IPFS
+- Stream
+- Object, with format `{ path: </foo/bar>, stream: <readable stream> }`
+- Array, an [array] of objects with the above format. 
+
+`options` is a optional argument of type object, that can contain the following properties:
+
+- `Recursive`. Add directory paths recursively.
+
+`callback` must follow `function (err, object) {}` signature, where `err` is an error if the operation was not successful and `object` is an ndjson return with properties { Name (string), Hash (multihash) } 
+
+If no `callback` is passed, a promise is returned.
+
+
+
+
+
+### `Cat` 
+
+> Displays the data contained by an IPFS object(s) at the given path..
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.cat(multihash, [callback])
+
+`multihash` is a [multihash]() which can be passed as:
+
+- Buffer, the raw Buffer of the multihash (or of and encoded version)
+- String, the toString version of the multihash (or of an encoded version)
+
+`callback` must follow `function (err, stream) {}` signature, where `err` is an error if the operation was not successful and `stream` is a readable stream.
+
+If no `callback` is passed, a promise is returned.
+
+
+
+
+
+### `Get` 
+
+> Download IPFS UnixFS files and directories with tar archiving and gzip compression options.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.get(multihash, [options, callback])
+
+`multihash` is a [multihash]() which can be passed as:
+
+- Buffer, the raw Buffer of the multihash (or of and encoded version)
+- String, the toString version of the multihash (or of an encoded version)
+
+`options` is a optional argument of type object, that can contain the following properties:
+
+- `output`, string - The path where output should be stored.
+- `archive`, bool   - Output a TAR archive.
+- `compress`, bool   - Compress the output with GZIP compression.
+- `compression-level` int    - The level of compression (1-9).
+
+`callback` must follow `function (err, ee) {}` signature, where `err` is an error if the operation was not successful and `ee` is an event emitter containing an object of {stream, path, dir} where dir is a boolean flag for a directory path. [ee](https://github.com/ipfs/js-ipfs-unixfs-engine/blob/master/src/exporter.js) // This should be upgraded to duplex object stream.
+
+If no `callback` is passed, a promise is returned.
+
+
+
+
+
 ## Object
 
 ### `object.new`


README.md
```diff
@@ -54,6 +54,83 @@ test.all(common)
...
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.add(path, [options, callback])
+
+`path` The path to a file to be added to IPFS. Can of type:
+
+- Path, String format path to a file to be added to IPFS
+- Object, with format `{ path: </foo/bar>, stream: <data> }`
+- Array, an [array] of objects with the above format. 
+
```
diasdavid: there is also the option that no 1st arg is passed, a duplex stream is returned

README.md
```diff
@@ -54,6 +54,83 @@ test.all(common)
...
+- String, the toString version of the multihash (or of an encoded version)
+
+`options` is a optional argument of type object, that can contain the following properties:
+
+- `output`, string - The path where output should be stored.
+- `archive`, bool   - Output a TAR archive.
+- `compress`, bool   - Compress the output with GZIP compression.
+- `compression-level` int    - The level of compression (1-9).
+
+`callback` must follow `function (err, ee) {}` signature, where `err` is an error if the operation was not successful and `ee` is an event emitter containing an object of {stream, path, dir} where dir is a boolean flag for a directory path. [ee](https://github.com/ipfs/js-ipfs-unixfs-engine/blob/master/src/exporter.js) // This should be upgraded to duplex object stream.
```
diasdavid: Since we are moving and defining the spec. Let's move the 'exporter' to become a duplex stream instead of an event emitter, otherwise it will be a lil strange why one is EE and another is DuplexStream. This way we will be also able to leverage the streams properties.

diasdavid:
This is s great start! Thank you @nginnever :D

README.md
nginnever: Completely agree!

README.md
nginnever: Ahh good, added

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+- Path, String format path to a file to be added to IPFS
+- Stream
+- Object, with format `{ path: </foo/bar>, stream: <readable stream> }`
+- Array, an [array] of objects with the above format. 
+
+`options` is a optional argument of type object, that can contain the following properties:
+
+- `Recursive`. Add directory paths recursively.
+
+`callback` must follow `function (err, object) {}` signature, where `err` is an error if the operation was not successful and `object` is an ndjson return with properties { Name (string), Hash (multihash) }
```
noffle: What is "an ndjson return"?

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+
+> Store a Unixfs file or directory of files .
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.add(path, [options, callback])
+
+`path` The path to a file to be added to IPFS. Can be of type:
+
+- Callback, If no first argument is supplied a callback will return a duplex stream.
```
noffle: What are the semantics of this duplex stream?

noffle: s/Callback/function

noffle: "If no first argument is supplied" or "if a function is supplied"?

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+- Path, String format path to a file to be added to IPFS
```
noffle: "string, in the form of a filesystem path to a file to be added"

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+- Stream
```
noffle: what kind? readable? writeable? transform? duplex?

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+- Object, with format `{ path: </foo/bar>, stream: <readable stream> }`
+- Array, an [array] of objects with the above format. 
+
+`options` is a optional argument of type object, that can contain the following properties:
+
+- `Recursive`. Add directory paths recursively.
```
noffle: can we do s/Recursive/recursive? it's js idiom to use camelcase names

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+
+### `Add`
+
+> Store a Unixfs file or directory of files .
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.add(path, [options, callback])
+
+`path` The path to a file to be added to IPFS. Can be of type:
```
noffle: maybe "toAdd" is a better name than "path", since it can take on many values; many of which are not paths

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
 
 A valid (read: that follows this interface) IPFS core implementation, must expose the following API.
 
+## Group Files
+
+### `Add`
+
+> Store a Unixfs file or directory of files .
```
noffle: can we either explain what "unixfs" means or link to a document that does?

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
 
 A valid (read: that follows this interface) IPFS core implementation, must expose the following API.
 
+## Group Files
```
noffle: Is "Group Files" a canonical api section name? I haven't seen this referred to anywhere -- could you link me?

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+
+If no `callback` is passed, a promise is returned.
+
+
+
+
+
+### `Cat` 
+
+> Displays the data contained by an IPFS object(s) at the given path..
```
noffle: Does it "display" an object's data or does it retrieve it? Maybe "streams an IPFS object's data, given an IPFS multihash"?

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.cat(multihash, [callback])
+
+`multihash` is a [multihash]() which can be passed as:
```
noffle: `s/[multihash]()/[multihash][]` so we can take advantage of markdown footnotes (I added one for `multihash` in a recent PR)

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+##### `JavaScript` - ipfs.get(multihash, [options, callback])
+
+`multihash` is a [multihash]() which can be passed as:
+
+- Buffer, the raw Buffer of the multihash (or of and encoded version)
+- String, the toString version of the multihash (or of an encoded version)
+
+`options` is a optional argument of type object, that can contain the following properties:
+
+- `output`, string - The path where output should be stored.
```
noffle: "... on the local file system"

noffle: what are the semantics if this is vs is not set? If set, does this give us the below duplex stream **and** write to the local fs? If not set, does it not write?

README.md
```diff
@@ -54,6 +54,85 @@ test.all(common)
...
+- `archive`, bool   - Output a TAR archive.
```
noffle: for each file? for the entire file+directory structure?

noffle:
Sorry for the million line comments! :grinning:

README.md
diasdavid: unixfs is all lowercase

README.md
diasdavid: or even call it `files`

README.md
diasdavid: This is not correct. If no callback is passed, then a object duplex stream is returned.

README.md
nginnever: readable, ty! unless duplex works too pipe data to the through fixed sized chunker https://github.com/ipfs/js-ipfs-unixfs-engine/blob/master/src/importer.js#L67

README.md
nginnever: Group is used it seems for the top level commands in the apiary, and group files is already taken from the mfs group, nomenclature ftw! we should talk more on what we call commands in ipfs in general.

README.md
nginnever: according to go-ipfs ```ipfs cat --help``` it "displays" the data

README.md
nginnever: No output string creates files on the cwd for the cli. Not sure how this will be handled exactly yet if provided to the http-api, but that is a different spec and this probably belongs there?

README.md
nginnever: have not yet figured that one out yet :D

README.md
noffle: even `files` is kind of misleading, since a stream !== a file. heck, maybe even something as nebulous as `data`. the real problem here is that this api call is becoming a kitchen sink; it becomes hard to define concrete terminology when you accept so many different things in one place.

README.md
noffle: is this just an apiary thing, or does putting the word "group" in front of everything add extra value to the reader?

README.md
noffle: this seems a bit confusing. good layering would be to have the core just provide plain streams to the caller (idempotent; doesn't mutate global fs), and let downstream (e.g. cli) implement things like fs writes.

nginnever: agreed, think i was trying to add documentation from what the cli/http server would handle the streams. Core will simply be streams.

README.md
dignifiedquire: There is no need for a duplex stream on `get`. This should be a readable stream

nginnever: As long as the stream can stay an object stream for getting paths back with the stream, 

https://github.com/ipfs/js-ipfs-unixfs-engine/blob/master/src/exporter.js#L55

diasdavid:
There is a lot of great activity and we are really close of having add, cat and get in both js-ipfs and js-ipfs-api with the same API, which is great.

@noffle @nginnever, let's set this as a priority for the day, as we need ipfs cat to unlock @haadcode.

What is missing

- [ ] Writing the tests for the described API in this PR (writing the tests is mostly reusing the existing test codebase available in js-ipfs and js-ipfs-api
- [ ] merge and released unixfs-engine with new exporter API https://github.com/ipfs/js-ipfs-unixfs-engine/pull/42
- [ ] merge and release js-ipfs with ipfs.cat https://github.com/ipfs/js-ipfs/pull/227
- [ ] merge and release js-ipfs with ipfs.add https://github.com/ipfs/js-ipfs/pull/235
- [ ] merge and release js-ipfs with ipfs.get https://github.com/ipfs/js-ipfs/pull/205
- [ ] merge and release js-ipfs-api with ipfs.get https://github.com/ipfs/js-ipfs-api/pull/271

diasdavid:
@nginnever is this PR still up to date?

nginnever:
hey @diasdavid, no this os dated and should be closed. 

```get``` spec is the only thing missing but has moved from the PR's on this list.

diasdavid:
@nginnever thank you for confirming :)

PR #15, opened 2016-05-17
draft api docs for blocks

noffle:
1. `del` was intentionally omitted. It's not in js-ipfs-api, so let's not commit to it until there's a clear need / demand.
2. js-ipfs-api and js-ipfs core disagree on the return object of `block.stat`. The former is *way* more descriptive than the latter, so we can go with that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,82 @@ test.all(common)
 
 A valid (read: that follows this interface) IPFS core implementation, must expose the following API.
 
+## Block
+
+### `block.put`
+
+> Create a new IPFS Block.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.block.put(data, [callback])
+
+Where `data` can be a
+
+- Buffer, requiring that the encoding is specified on the options. If no
+  encoding is specified, Buffer is treated as the Data field
+- [Block][] instance
+
+`callback` has the signature `function (err, block) {}`, where `err` is an error
+if the operation was not successful. and `block` is a [Block][].
+
+
+
+
+### `block.get`
+
+> Retrieve an IPFS Block from the underlying Block Store by its mulithash.
+
+##### `Go` **WIP**
+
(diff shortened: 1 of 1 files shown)


diasdavid:
> js-ipfs-api and js-ipfs core disagree on the return object of block.stat. The former is way more descriptive than the latter, so we can go with that.

Can we provide the same level of information in js-ipfs-api?

noffle:
Good catch! I looked and no -- no we can not. go-ipfs doesn't provide more than `Key` and `Size`, so we'll need to just provide those.

diasdavid:
we can probably shim that by making one more call?

noffle:
LGTM to merge this?

diasdavid:
Oh, let's not merge without tests and the tests being fully applied to both js-ipfs and js-ipfs-api. We don't want to expose a interface that isn't being used yet. We want to make sure that this document can be used as reference for implementation and for usage.

diasdavid:
Also, another thing is that we don't want to nail down a spec for API without getting a feeling for it during implementation process.

//cc @dignifiedquire @nginnever

dignifiedquire:
Done in https://github.com/ipfs/interface-ipfs-core/pull/16

diasdavid:
Thank you @dignifiedquire. @noffle I believe you have open a PR again

noffle:
I'm in favour of getting docs merged once they are LGTM'd to make it clear they're accepted, and then work on the impl in a separate PR. We can make intentions clear at the top of the readme that this is very WIP / pre-release until then.

diasdavid:
Somethings result from the experience of actually making the tests and adapting the API. Writing a spec and not testing it will lead to trouble

noffle:
Accepting a PR into master is not analogous to it being set in stone -- especially during early development. It gives us an agreed upon vector, with the knowledge that it can change as needed.

noffle:
Not worth further bikeshedding on -- let's just get the impl in. :smile:

PR #17, opened 2016-05-18
Fix README typo + tweaks.

noffle:
s/interface-stream-muxer/interface-ipfs-core, and a couple of other small tweaks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,15 +12,16 @@ The API is presented with both Node.js and Go primitives. However, there are no 
 
 # Modules that implement the interface
 
-- [JavaScript IPFS implementation](https://github.com/ipfs/js-ipfs)
-- [JavaScript ipfs-api](https://github.com/ipfs/js-ipfs-api)
+- **WIP** [JavaScript IPFS implementation](https://github.com/ipfs/js-ipfs)
+- **WIP** [JavaScript ipfs-api](https://github.com/ipfs/js-ipfs-api)
 - Soon, go-ipfs, go-ipfs-api, java-ipfs-api, python-ipfs-api and others will implement it as well.
 
-Send a PR to add a new one if you happen to find or write one.
+Send in a PR if you find or write one!
 
 # Badge
 
-Include this badge in your readme if you make a new module that uses interface-stream-muxer API.
+Include this badge in your readme if you make a new module that implements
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,7 +67,7 @@ A valid (read: that follows this interface) IPFS core implementation, must expos
 
 `callback` must follow `function (err, node) {}` signature, where `err` is an error if the operation was not successful and `node` is a MerkleDAG node of the type [DAGNode][]
 
-If no `callback` is passed, a promise is returned.
+Returns a [promise][].
```
dignifiedquire: This change is not correct. The promise is only returned if no `callback` is passed.

noffle: Alas! I thought so too, but tried this with `promisify-es6` -- [it returns a promise regardless](https://gist.github.com/noffle/aac170fc16890c77809f0ba14d9fa235).

dignifiedquire: Interesting. Not sure if we want this though, I guess we have to fork it and make our own version in the long run.

RichardLitt:
I'm liking the aggressive linking you're doing recently. +1.

README.md
noffle: @dignifiedquire I think I agree about it being undesirable to always return a promise. A counterargument though might be that *lots* of people are already depending on promisify and us deviating from "the norm" may be detrimental to the larger community. Hm--I'm not sure. ping @diasdavid @nginnever too :)

noffle:
@RichardLitt man, ever since I learned about markdown footnote links they've soo been my jam :heart:

README.md
diasdavid: I would rather for it to be one or the other. Why can't we reach out to the dev of promisify-es6 and understand the reason behind his decision? Seems like the dev might have expected that if someone passes a callback, to not grab the promise as well, there would be nothing good coming from that.

diasdavid: Nevertheless, as for the spec. Let's rephrase to:

`A promise is also returned, you can skip passing the callback to the function.`

noffle: +1 for reaching out to the dev first -- this is how open source works best!

noffle:
I reverted the phrasing back to how it was before for promises, but linkified them.

diasdavid:
thank you @noffle :)

PR #20, opened 2016-05-20
API proposal for ipfs.files.add

noffle:
This is my first draft at merging the wildly different js-ipfs-api and js-ipfs `add` APIs. Notable things:

1. I dropped support for adding files from the local filesystem and for adding urls. I've written up [a plan for making this change](https://github.com/ipfs/js-ipfs-api/issues/279). These are nice utilities, but they don't belong in core.
2. Both object tuple formats `(path, stream)` *and* `(path, content)` are included here. It'd be nice if we could merge them into one style though. Maybe if `content` could be a Buffer *or* Readable?
3. I made the return object format consistent with what js-ipfs already does. This will require [a change](https://github.com/ipfs/js-ipfs/issues/247) to js-ipfs or unixfs-engine.
4. We also accept a Buffer or a Readable stream.
5. If no argument is passed in, a Writeable stream is returned. I feel mixed about this. I understand the convenience of being able to pipe directly into `ipfs.files.add()`, but this style is confusing in the way it overloads the API call. I'm on the preference of dropping it but letting folks develop their own utilities for it if they're so inclined. This also gets ambiguous when I don't provide `data` **or** `callback` -- do I get back a Promise OR a Writeable? :D

---

ping @diasdavid @nginnever @dignifiedquire


haadcode:
I would like the add to be:
`ipfs.files.add('/tmp/myfile.txt')` --> 'Qm...Foo'.

It'd be also fine to return a `DAGNode`, ie `ipfs.files.add(path).toJSON().Hash` --> 'Qm...Foo`.

 If we want to give fine-grained options to define how/what the data is, being able to pass it as @noffle suggested sounds good.

There's couple of reasons why:
- ipfs.files.add(path) is consistent with go-ipfs CLI and does exactly what I would expect it to do
- As a user, I shouldn't have to care how the file is read on the system level, I'm hooking up to a *file system* so it should function as such and abstract the system level functionality away from me
- `ipfs.files.addFiles(path)` adds redundancy in the naming. It'd work better as `ipfs.addFiles(path)`. Or perhaps as convenience method like: `ipfs.files.addDirectory(path)`.
- Currently `ipfs.add(file)` returns the full path of the file as an array of hashes and imo this is undesirable default behaviour. If a full path is needed, it should be an option, like `recursive` is now.

README.md
```diff
@@ -55,6 +55,55 @@ test.all(common)
...
+```js
+{
+  path: '/tmp/myfile.txt',
+  content: new Buffer('hello')
+}
+```
+- a `Buffer` instance
+- a readable stream
+
+If no value `data` is provided, a Readable stream will be returned, which
```
diasdavid: s/Readable/Duplex

README.md
```diff
@@ -55,6 +55,55 @@ test.all(common)
...
+error if the operation was not successful and `res` is an array of objects of
+the form
+
+```js
+// js-ipfs-api
+{
+  Name: 'some name',
+  Hash: 'QmRNjDeKStKGTQXnJ2NFqeQ9oW23WcpbmvCVrpDHgDg3T6'
+}
+```
```
diasdavid: https://github.com/ipfs/js-ipfs/issues/247#issuecomment-220767561

noffle:
Thanks for weighing in, @haadcode! Responses:

> I would like the add to be:
> `ipfs.files.add('/tmp/myfile.txt')` --> 'Qm...Foo'.

We're really keen on having IPFS Core only expose basic operations that all IPFS nodes will be able to support. Adding files from the local filesystem is something that e.g. the browser can't do, so I think it doesn't belong in core. That's *not* to say however that APIs on top couldn't add this functionality!

> It'd be also fine to return a `DAGNode`, ie `ipfs.files.add(path).toJSON().Hash` --> 'Qm...Foo`.

Pinging @diasdavid for thoughts on returning first-class objects vs JSON -- I'm not sure what I think yet.

> - `ipfs.files.add(path)` is consistent with go-ipfs CLI and does exactly what I would expect it to do

I think we set up a bad expectation here: that CLI should be 1:1 with Core. CLI is a *specific interface* into the broader world of Core, since CLI has access to things like the local FS, pipes, a shell interpreter, etc.  Because of this Core cannot (and shouldn't) match CLI's capabilities. In the case of `ipfs add`ing a file, I think it makes more sense for the CLI to do the glue work of turning the files/dirs into streams and passing them into Core. I think @nginnever did this in unixfs-engine importer?

> - As a user, I shouldn't have to care how the file is read on the system level, I'm hooking up to a *file system* so it should function as such and abstract the system level functionality away from me

How should the API abstract this on the browser, though? Treat file paths as index-db paths? What about future IPFS nodes that have no storage like this at all?

> - `ipfs.files.addFiles(path)` adds redundancy in the naming. It'd work better as `ipfs.addFiles(path)`. Or perhaps as convenience method like: `ipfs.files.addDirectory(path)`.

Sure! I'm down for either. :)

> - Currently `ipfs.add(file)` returns the full path of the file as an array of hashes and imo this is undesirable default behaviour. If a full path is needed, it should be an option, like `recursive` is now.

Sorry, could you try to explain this again with some more context (maybe an example)? I'm not sure I understand.

README.md
noffle: Oops -- I meant Writable. What's the use of this being a Duplex here?

README.md
noffle: https://github.com/ipfs/js-ipfs/issues/247#issuecomment-220787339 :grinning:

README.md
diasdavid: You need to get back the DAGNodes correspondent to the files (and directories) that you are adding.

diasdavid:
`ipfs-core` won't make reads (or writes) on the filesystem. Files are passed to ipfs-core, by application code or through the http-api, if using the daemon. This is the same pattern used in go-ipfs.

README.md
```diff
@@ -55,6 +55,55 @@ test.all(common)
...
+
+If no `callback` is passed, a promise is returned.
+
+
```
diasdavid: If no `data` is passed, then, a duplex stream is returned, that pretty much is the duplex stream returned by unixfs-engine Importer

diasdavid:
👍 for making the returned objects be first class DAGNodes :)

README.md
diasdavid: This is related with having it returning a full DAGNode (https://github.com/ipfs/interface-ipfs-core/pull/20#issuecomment-220788843), which from what I read you also agree, right?

README.md
noffle: Ah right, it's a transform that's happening *over the API*! Got it. Thanks. :)

README.md
noffle: What if no `data` is passed *and* no `callback` is passed? The possible return matrix starts to get messy -- you'd want a stream *and* a promise back. ;)

diasdavid: That would be a misuse of the API.

Note that we need the ability to return a duplex stream to be able to 'add files as they come', through the http-api or through the cli.

README.md
noffle: Yes, I would like DAGNodes returned, but doesn't it come at a high cost? I.e. Won't doing an extra `object.get` call per added node mean N extra API requests against Core (in js-ipfs-api land, N HTTP requests)? Or is there some sort of batching `object.get` mechanism I don't know about?

README.md
noffle: We can have the callback/promise provide a stream if no `data` is passed, but we can't have the function actually return a stream ever, since we went down the road of always returning promises.

noffle:
Revisions applied -- just our convo re DAGNodes to resolve.

(Aside: I really wish we could simplify this API method: 4 possible input types, 2 types of return values, and both callback and promise control flows.)

README.md
diasdavid: Let's return the duplex stream on the callback on the interface-ipfs-core. This does not require changes on unixfs-engine. Sounds good?

README.md
diasdavid: The 'high cost' is virtual, because currently it is a given fact that the http-api needs improvements and it is better to provide a richer and consistent interface at a cost of another RTT, feeding back that into the http-api design, that simply cut it off.

noffle: Revision made.

README.md
```diff
@@ -55,6 +55,46 @@ test.all(common)
...
+}
+```
+- a `Buffer` instance
+- a Readable stream
+
+`callback` must follow `function (err, res) {}` signature, where `err` is an
+error if the operation was not successful. If no value `data` is provided, `res`
+will be a Duplex stream, to which tuples like the above two object formats can
+be written and [DAGNode][] objects will be outputted. Otherwise, `res` will be
+an array of [DAGNode][]s.
```
diasdavid: Give an example please.

```
stream.write({path: <path to file>, stream: <readable stream>})
// write as many as you want
stream.end()
stream.on('data', function (fileDAGNode) {
})
```

diasdavid:
Good progress :) Let's have tests :)

README.md
noffle: Added.

noffle:
- [ ] js-ipfs: https://github.com/ipfs/js-ipfs/pull/261
- [ ] js-ipfs-api: https://github.com/ipfs/js-ipfs-api/pull/281

README.md
```diff
@@ -55,6 +55,73 @@ test.all(common)
...
+
+If no `callback` is passed, a promise is returned.
```
haadcode: What does the Promise resolve to? An array of DAGNodes?

We should add examples for the returns
```javascript
ipfs.files.add(path).then((res) => console.log(res)) // --> [DAGNode1, DAGNode2, ...] ???
```

haadcode:
Good progress! 

My thoughts atm:
- I still think we really shouldn't limit the input arguments to `an array of objects, each of the form...`. We should be able to pass a path (or an array of paths). I understand the argument that the browsers don't have a file system but we also have the Node.js version *and* js-ipfs-api that both can support it and on browser, you can add files with the array of objects structure and have control over where the "file" data comes from. It's a lot more intuitive: `ipfs, add some *files*, from this path` as opposed to `ipfs, add some *data*, that you should handle as files, and they come in like this...`.
- I agree different *types* of result values is not ideal. However a callback is not a return value and as such should be fine if we always **return** a Promise and the callback's result argument is *always* a stream. This way the user has access to both: an actual result (array from the Promise) or a stream for further processing.
- We should try to combine the input argument's object notation to contain only one type of object: `{ path: '/path', content: <buffer or stream> }` and let the implementation handle the check which one it is.

diasdavid:
> I still think we really shouldn't limit the input arguments to an array of objects, each of the form.... We should be able to pass a path (or an array of paths). I understand the argument that the browsers don't have a file system but we also have the Node.js version and js-ipfs-api that both can support it and on browser, you can add files with the array of objects structure and have control over where the "file" data comes from. It's a lot more intuitive: ipfs, add some *files*, from this path as opposed to ipfs, add some *data*, that you should handle as files, and they come in like this....

The result from the discussion we had (although I see not document) is that there will be, inside js-ipfs-api, extra API calls like:

`ipfs.cli.add` - which takes an input path, just like the CLI

These calls will only be available on the js-ipfs-api and are not part of the interface-core-spec

> I agree different types of result values is not ideal. However a callback is not a return value and as such should be fine if we always return a Promise and the callback's result argument is always a stream. This way the user has access to both: an actual result (array from the Promise) or a stream for further processing.

Agree. The 'return value' (aka result) should be (and is) always the same, in this case. It should be the stream that will emit the several DAGNodes. 

> We should try to combine the input argument's object notation to contain only one type of object: { path: '/path', content: <buffer or stream> } and let the implementation handle the check which one it is.

Agree and I think this is what it is at the moment, did I miss something?

haadcode:
> ipfs.cli.add - which takes an input path, just like the CLI

Will this be available in Node.js version of js-ipfs? I think it should.

> These calls will only be available on the js-ipfs-api and are not part of the interface-core-spec

I think the namespacing here is what makes it confusing for me. If we move the add under `ipfs.cli.add` and regardless have `ipfs.files.add`, people will use the latter as it makes sense intuitively whereas its input arguments imo don't. I think there might be a conflict of perspectives here: I'm looking at it from a higher-level, app developer perspective and perhaps what is proposed here is akin more to a library/module developer perspective. While both important, the top-level API needs to have as simple and intuitive UX as possible. The way I read it atm, with the proposed api, is that I'm adding arbitrary *data* to ipfs via the files api, not *files*. A file has a path, a file is located somewhere, so the intuitive thinking says I need to tell the api which file it is I want to add. Do you see what I mean? 

> Agree. The 'return value' (aka result) should be (and is) always the same, in this case. It should be the stream that will emit the several DAGNodes.

Wait, what's your understanding of the value the returned Promise resolves to? A stream or an array?

> We should try to combine the input argument's object notation to contain only one type of object: { path: '/path', content: } and let the implementation handle the check which one it is.
> Agree and I think this is what it is at the moment, did I miss something?

It's not reflected (yet?) in the [PR](https://github.com/ipfs/interface-ipfs-core/pull/20/files), perhaps I'm looking at the wrong PR?

haadcode:
Discussion today:
```
dignifiedquire> I think it would be better to have two different methods, than to try and shoehorn everything into one
daviddias> I'm also happy to have two different methods
dignifiedquire> just have a method `createAddStream`
dignifiedquire> then everybody knows what to expect
haad> was thinking the same.
daviddias> Ok, let's do that.
daviddias> Can someone type that while I'm at the subway? So that we don't loose this decision? :)
```

This sounds really good to me. Separate the stream and Promise control flows and it'll simplify everything nicely.

noffle:
Cool! Lots of thinking happened while I was sleeping.

Yes: let's have functions that do *one single thing* than be a grab bag of inputs and outputs! Here's what I've aggregated from your comments:

1. break `add` into `add` and `createAddStream` (@dignifiedquire)
2. have one form of tuple input (`{ path, content }`) rather than supporting `stream` and `content` keys (@haadcode)

I've made these revisions here.

---

@haadcode and @diasdavid re `ipfs.cli.add` etc for file adding: I think the point of confusion here is on what `interface-ipfs-core` is providing. Here are two things this module is:

1. an interface for core ipfs functions
2. a way to use either a local JS node or a remote go/js node over HTTP with a single interface

Here is something *it is not*:

1. a drop-in replacement for all of the functionality in js-ipfs-api

It's unfortunate, but js-ipfs-api does more things than what "Core IPFS"  does, so that overflow needs to go somewhere in its API.

So, and niceties that js-ipfs-api provides (like reading files from the FS) needs to be exposed from a higher level API than Core. Maybe this means a new module that builds atop `interface-ipfs-core`, like `js-node-ipfs-helper` that can do all of these local FS helper operations.

dignifiedquire:
Re part 1: Yes that was what we meant 

---

Re part 2: 

I think it's a really good notion of having a module that wraps around both `ipfs-core` as well as `js-ipfs-api` to provide additional functionality, like interacting with the file system.

test/files.js
```diff
@@ -0,0 +1,154 @@
...
+
+    it('add a nested dir as array', (done) => {
+      if (!isNode) return done()
+      const fs = require('fs')
+      const base = path.join(__dirname, 'data/test-folder')
+      const content = (name) => ({
+        path: `test-folder/${name}`,
+        content: fs.readFileSync(path.join(base, name))
+      })
+      const dirs = [
```
nginnever: It would be nice if there was a test for an empty directory by adding one in ```data/test-folder/empty``` so that it can check <a href="https://github.com/ipfs/js-ipfs/pull/256/commits/3b8f4b6448febb0c84aaecb0f111f9001d31b10a">this addition</a> to js-ipfs pr.

test/files.js
```diff
@@ -0,0 +1,154 @@
...
+    describe('promise', () => {
+      it('buffer', () => {
+        let buf = new Buffer(testfile)
+        return ipfs.files.add(buf)
+          .then((res) => {
+            const added = res[0] != null ? res[0] : res
+            const mh = bs58.encode(added.multihash()).toString()
+            expect(mh).to.equal('Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP')
+            expect(added.links).to.have.length(0)
+          })
```
diasdavid: Missing `.catch(err)`

README.md
```diff
@@ -55,6 +55,86 @@ test.all(common)
...
+
+### `createAddStream`
+
+> Add files and data to IPFS using a transform stream.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.files.createAddStream([callback])
+
+Provides a Transform stream, where objects can be written of the forms
```
diasdavid: Provides a `Duplex` stream

README.md
```diff
@@ -55,6 +55,86 @@ test.all(common)
...
+
+```js
+{
+  path: '/tmp/myfile.txt',
+  content: (Buffer or Readable stream)
+}
+```
+
+`callback` must follow `function (err, stream) {}` signature, where `err` is an
+error if the operation was not successful. `stream` will be a Transform stream,
```
diasdavid: `Duplex stream`

README.md
```diff
@@ -55,6 +55,86 @@ test.all(common)
...
+to which tuples like the above two object formats can be written and [DAGNode][]
+objects will be outputted.
+
+If no `callback` is passed, a promise is returned.
+
+```js
+ipfs.files.createAddStream(function (err, stream) {
+  stream.on('data', function (dagNode) {
```
diasdavid: We can't loose the path here. It will be important information, since the DAGNode doesn't have the path that someone uses to point to it.
The callback should have the following format:
```
function (file) {
  console.log(file.path)
  console.log(file.node.multihash()
}
```

package.json
```diff
@@ -28,9 +28,12 @@
   },
   "homepage": "https://github.com/ipfs/interface-ipfs-core#readme",
   "dependencies": {
+    "bl": "^1.1.2",
     "bs58": "^3.0.0",
     "chai": "^3.5.0",
-    "ipfs-merkle-dag": "^0.6.0"
+    "detect-node": "^2.0.3",
+    "ipfs-merkle-dag": "^0.6.0",
+    "readable-stream": "^2.1.4"
```
diasdavid: We have to use ` "readable-stream": "^1.1.13",` cause version two introduced a issue when browserifying the code

https://github.com/ipfs/js-ipfs-unixfs-engine/blob/master/package.json#L59

test/data/test-folder/add.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+})
+
+ipfs.add(['./files/hello.txt', './files/ipfs.txt'], function (err, res) {
+  if (err || !res) return console.log(err)
+
+  for (let i = 0; i < res.length; i++) {
+    console.log(res[i])
+  }
+})
```
diasdavid: Should these tests be here? It would be better one example of file to add that is not code.

test/data/test-folder/cat.js
```diff
@@ -0,0 +1,18 @@
...
+
+ipfs.cat(hash, function (err, res) {
+  if (err || !res) return console.log(err)
+
+  if (res.readable) {
+    res.pipe(process.stdout)
+  } else {
+    console.log(res)
+  }
+})
```
diasdavid: same as above

test/data/test-folder/ipfs-add.js
```diff
@@ -0,0 +1,14 @@
...
+const ipfs = require('../src')('localhost', 5001)
+const files = process.argv.slice(2)
+
+ipfs.add(files, {recursive: true}, function (err, res) {
+  if (err || !res) return console.log(err)
+
+  for (let i = 0; i < res.length; i++) {
+    console.log('added', res[i].Hash, res[i].Name)
+  }
+})
```
diasdavid: same as above

test/data/test-folder/ls.js
```diff
@@ -0,0 +1,18 @@
...
+
+  res.Objects.forEach(function (node) {
+    console.log(node.Hash)
+
+    console.log('Links [%d]', node.Links.length)
+    node.Links.forEach(function (link, i) {
+      console.log('[%d]', i, link)
+    })
+  })
+})
```
diasdavid: same as above

test/data/test-folder/version.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
+
+const ipfs = require('../src')('localhost', 5001)
+
+ipfs.commands(function (err, res) {
+  if (err) throw err
+  console.log(res)
+})
```
diasdavid: same as before

diasdavid:
The tests should cover adding a files:

- [ ] file with less than a block size
- [ ] file bigger than a block size
- [ ] file bigger than a block size, but that is not a multiple of the block size (size % 256KiB !== 0)
- [ ] file that is 'big' (between 10 and 20Mb)
- [ ] directory with files
- [ ] directory with files and empty dirs
- [ ] directory with other directories that have files and other empty dirs (nested dirs)

diasdavid:
_Note for the future:_ @noffle open branches on the main repos, so that others can collaborate without having to PR your PR in order to make a change :)

package.json
dignifiedquire: it has to be `"1.1.3"` because the issue was backported to `"1.1.14"`

test/data/test-folder/version.js
noffle: yup, I got it from the one time :)

noffle:
@diasdavid I filed https://github.com/ipfs/interface-ipfs-core/issues/23 for the tests you mention. I'm going to focus on getting this code in before expanding the test suite.

noffle:
> open branches on the main repos

I get used to the general github workflow, where you don't tend to have `origin` write access. :grinning:

test/files.js
noffle: Good idea. Delegating this to https://github.com/ipfs/interface-ipfs-core/issues/23

README.md
noffle: I spent some time thinking about this, and I'm convinced that it's actually a `Transform` stream. [Here](http://stackoverflow.com/questions/18335499/nodejs-whats-the-difference-between-a-duplex-stream-and-a-transform-stream/18339203#18339203) is the article that helped me draw a clearer line of differentiation between the two.

It's a Transform because the input and the output are directly related: each object written is outputted as an object that's been transformed (/w side effects) through the IPFS API.

README.md
noffle: see above

package.json
noffle: Didn't realize this -- thanks!

test/data/test-folder/add.js
noffle: You're right: this is confusing.

noffle: Project Gutenberg to the rescue! :heart:

test/data/test-folder/cat.js
noffle: done

test/data/test-folder/ipfs-add.js
noffle: done

test/data/test-folder/ls.js
noffle: done

test/data/test-folder/version.js
noffle: done

README.md
noffle: Agreed! This is really important. I'll add this and update the spec.

README.md
diasdavid: There are more objects emitted than those who are written. E.g

`/a/b/c.txt`

will emit 3 times, one for the file and 2 other for the dirs.

I understand your reasoning, but feels weird to call it a Transform, it would be like calling a TCP socket a Transform stream, since we will read mutated output. I think both will work

@dignifiedquire @haadcode thoughts?

noffle: Transform streams actually very commonly emit more or fewer elements
than they are given: a zlib compress transform will yield fewer or more
bytes depending on direction; a [binary fsk
encoder](https://github.com/noffle/binary-fsk) will accept byte buffer
objects but emit *way* more audio data depending on the sample rate.

The key distinction is that in Transform streams, the output is in some
way calculated from the input. This is in contrast to a Duplex stream
(e.g. TCP), which is much more akin to two people on a telephone
exchanging messages.

dignifiedquire: A transform is stream is just a specialised version of a duplex stream, so really both are correct.

noffle:
All feedback has been addressed! This is ready for merge if all LGTY, @diasdavid.

diasdavid:
Awesome work! SO CLOSE :D 

It feels to me that [postponing the tests](https://github.com/ipfs/interface-ipfs-core/issues/23) that will check that this API operates as expect, might be a decision that will quickly backfire.

Anything against pushing this tests ahead? All of them are almost created over js-ipfs, js-ipfs-api and js-ipfs-unixfs-engine already, just need to be ported here.

README.md
```diff
@@ -55,6 +55,96 @@ test.all(common)
...
+
+```js
+{
+  path: '/tmp/myfile.txt',
+  node: DAGNode
+}
+```
+
+If no `callback` is passed, a promise is returned.
+
```
diasdavid: Add: `example:`

README.md
diasdavid: 👍

test/data/test-folder/alice.txt
```diff
@@ -0,0 +1,283 @@
...
+Lastly, she pictured to herself how this same little sister of hers
+would, in the after-time, be herself a grown woman; and how she would
+keep, through all her riper years, the simple and loving heart of her
+childhood: and how she would gather about her other little children, and
+make THEIR eyes bright and eager with many a strange tale, perhaps even
+with the dream of Wonderland of long ago: and how she would feel with
+all their simple sorrows, and find a pleasure in all their simple joys,
+remembering her own child-life, and the happy summer days.
+
+              THE END
```
diasdavid: Ahaha, love it :)

noffle:
@diasdavid done: these tests are all now present!

diasdavid:
👍 @noffle, rad :D 

Going to jump into a plane down, wanna squash the commits?

noffle:
Squashed and ready.

test/data/test-folder/ipfs-add.js
diasdavid: Do we still use this file?

test/files.js
```diff
@@ -0,0 +1,167 @@
...
+      if (isNode) {
+        const fs = require('fs')
+
+        const testfilePath = path.join(__dirname, './data/testfile.txt')
+        testfile = fs.readFileSync(testfilePath)
+
+        const testfileBigPath = path.join(__dirname, './data/15mb.random')
+        testfileBig = fs.createReadStream(testfileBigPath, { bufferSize: 128 })
+      } else {
+        testfile = require('raw!./data/testfile.txt')
```
diasdavid: We now use the browserify transform to support fs.readFileSync on the transpilation process as well. We don't need to use the `require('raw`

test/files.js
```diff
@@ -0,0 +1,167 @@
...
+    before((done) => {
+      // load test data
+      if (isNode) {
+        const fs = require('fs')
+
+        const testfilePath = path.join(__dirname, './data/testfile.txt')
+        testfile = fs.readFileSync(testfilePath)
+
+        const testfileBigPath = path.join(__dirname, './data/15mb.random')
+        testfileBig = fs.createReadStream(testfileBigPath, { bufferSize: 128 })
```
diasdavid: Does this mean that we don't test in the browser for a 15Mb buffer?

test/files.js
```diff
@@ -0,0 +1,167 @@
...
+        expect(res[0].path).to.equal('data.txt')
+        expect(res[0].node.size()).to.equal(17)
+        const mh = 'QmVv4Wz46JaZJeH5PMV4LGbRiiMKEmszPYY3g6fjGnVXBS'
+        expect(bs58.encode(res[0].node.multihash()).toString()).to.equal(mh)
+        done()
+      })
+    })
+
+    it('buffer as tuple', (done) => {
+      if (!isNode) return done()
```
diasdavid: Why not test it on the browser too?

test/files.js
```diff
@@ -0,0 +1,167 @@
...
+        expect(res[0].path).to.equal(mh)
+        expect(res[0].node.links).to.have.length(0)
+        done()
+      })
+    })
+
+    it('BIG buffer', (done) => {
+      if (!isNode) {
+        return done()
+      }
```
diasdavid: Why not test it on the browser? Any issue loading the 15Mb Buffer?

test/files.js
```diff
@@ -0,0 +1,167 @@
...
+        expect(res[0].node.links).to.have.length(58)
+        const mh = bs58.encode(res[0].node.multihash()).toString()
+        expect(res[0].path).to.equal(mh)
+        expect(mh).to.equal('Qme79tX2bViL26vNjPsF3DP1R9rMKMvnPYJiKTTKPrXJjq')
+        done()
+      })
+    })
+
+    it('add a nested dir as array', (done) => {
+      if (!isNode) return done()
```
diasdavid: same as above

noffle:
@diasdavid This is code I copied out of the existing repos, so these tests never ran in the browser previously either. Agreed 100% that we should make them do so, but given a) what work this PR is blocking, and b) that by not running in them in the browser we're no worse off than we were before (in fact, we're still better off, because there are most tests here than before!), I suggest we *merge this now* and come back to your those tests later. In fact, I've created [a tracking issue](https://github.com/ipfs/interface-ipfs-core/issues/25) for it. SGTM?

test/files.js
nginnever: Heyo! This is copied from <a href="https://github.com/ipfs/js-ipfs-api/blob/master/test/api/add.spec.js#L21">js-ipfs-api test spec</a>

Comment mentions browser failing with > 100mb in mem

test/files.js
nginnever: This should work in browser, will test when I get a moment

diasdavid:
Working on merging this. Made a branch on this repo so that we can all contribute to: https://github.com/ipfs/interface-ipfs-core/pull/26 , closing this one.

PR #21, opened 2016-05-21
Interface for `ipfs.files.cat` (and alias `ipfs.cat`)

nginnever:
This is the beginning of the cat spec. Some notable issues I can think of:

1) The test's aren't currently set up in npm package so I haven't tested the cat test file, but it is just a clone of the js-ipfs core test.
2) Not sure about the promises return. I have to understand promises a bit better or where they are currently being used. js-ipfs-api ```cat``` isn't set up with promises and the js-ipfs cli isn't either to my knowledge. cc @dignifiedquire

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,33 @@ ipfs.files.createAddStream(function (err, stream) {
 
 
 
+
+### `cat` 
+
+> Streams the file at the given IPFS multihash..
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.cat(multihash, [callback])
+
+`multihash` is a [multihash][] which can be passed as
+
+- Buffer, the raw Buffer of the multihash 
+- String, the base58 encoded version of the multihash
+
+`callback` must follow `function (err, stream) {}` signature, where `err` is an error if the operation was not successful and `stream` is a readable stream of the file.
+
+If no `callback` is passed, a promise is returned.
+
+```js
+ipfs.files.cat(multihash, function (err, file) {
+  // file will be a stream containing the data of the file requested
+})
+```
+
+
+
+
 ## Object
 
 ### `object.new`
--- a/src/files.js
+++ b/src/files.js
@@ -7,9 +7,10 @@ const Readable = require('readable-stream')
 const path = require('path')
 const fs = require('fs')
 const isNode = require('detect-node')
+const bl = require('bl')
 
 module.exports = (common) => {
-  describe('.files/add', () => {
+  describe('.files', () => {
     let smallFile
     let bigFile
     let ipfs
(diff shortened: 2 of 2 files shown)


diasdavid:
> 1) The test's aren't currently set up in npm package so I haven't tested the cat test file, but it is just a clone of the js-ipfs core test.

Couldn't you try them with npm link?

> 2) Not sure about the promises return. I have to understand promises a bit better or where they are currently being used. js-ipfs-api cat isn't set up with promises and the js-ipfs cli isn't either to my knowledge. cc @dignifiedquire

Every js-ipfs-api call has a promise API since last October/November. However, the previous set up doesn't work for how we now have to implement this cross compatible methods, but we do have a way to achieve that, which can be observed in the already implemented Object API - https://github.com/ipfs/js-ipfs-api/blob/master/src/api/object.js

README.md
```diff
@@ -55,6 +55,29 @@ test.all(common)
 
 A valid (read: that follows this interface) IPFS core implementation, must expose the following API.
 
+## Files
+
+### `Cat`
```
diasdavid: lowercase

test/files.js
```diff
@@ -0,0 +1,38 @@
...
+    after((done) => {
+      common.teardown(done)
+    })
+
+    describe('.cat', () => {
+      it('returns file stream', (done) => {
+        const hash = 'QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o'
+        ipfs.files.cat(hash, (err, res) => {
+          expect(err).to.not.exist
+          res.on('data', (data) => {
```
diasdavid: this is not matching the API description above. `res` should be the stream.

test/files.js
```diff
@@ -0,0 +1,38 @@
...
+            data.stream.pipe(bl((err, bldata) => {
+              expect(err).to.not.exist
+              expect(bldata.toString()).to.equal('hello world\n')
+              done()
+            }))
+          })
+        })
+      })
```
diasdavid: It needs more tests for the other variants of arguments, as you describe on the API description

test/files.js
```diff
@@ -0,0 +1,38 @@
...
+    })
+  })
```
diasdavid: Also misses the promises tests

diasdavid:
cat'ing a big file is important https://github.com/ipfs/js-ipfs-api/blob/master/test/api/cat.spec.js#L43-L58

test/files.js
nginnever:
```
ipfs.files.cat(hash, (err, stream) => {
         expect(err).to.not.exist
         stream.on('data', (file) => {
```

would be correct

diasdavid: not 'file'. `cat` is one file only, it should be the stream of the file directly.

README.md
```diff
@@ -55,6 +55,29 @@ test.all(common)
...
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.cat(multihash, [callback])
+
+`multihash` is a [multihash][] which can be passed as
+
+- Buffer, the raw Buffer of the multihash (or of and encoded version)
+- String, the toString version of the multihash (or of an encoded version)
+
+`callback` must follow `function (err, stream) {}` signature, where `err` is an error if the operation was not successful and `stream` is a readable stream in object mode that contains objects representing a file in the form of { path: fileName, stream: fileData }.
```
diasdavid: In cat there is no filename, just the stream of the content.


"`stream` is a readable stream of the content referenced by the multihash."

test/files.js
nginnever: I could make another test with 
```
const mhBuf = new Buffer(bs58.decode('Qm...'))
ipfs.files.cat(mhBuf, (err, stream) => {})
```

test/files.js
```diff
@@ -0,0 +1,38 @@
...
+      })
+    })
+
+    after((done) => {
+      common.teardown(done)
+    })
+
+    describe('.cat', () => {
+      it('returns file stream', (done) => {
+        const hash = 'QmT78zSuBmuS4z925WZfrqQ1qHaJ56DQaTfyMUF7F8ff5o'
```
diasdavid: right now you can use a hash for the 'init docs' that gets bundled with IPFS. But add a note saying to replace that hash by something you put with ipfs.add, once that is ready.

test/files.js
```diff
@@ -0,0 +1,38 @@
...
+              expect(err).to.not.exist
+              expect(bldata.toString()).to.equal('hello world\n')
+              done()
+            }))
+          })
+        })
+      })
+    })
+  })
+}
```
diasdavid: It would be good to also have a test for a non existing hash, so that we can also test for the err behaviour

README.md
nginnever: The core is returning a readable object stream from the exporter, which will send one object containing ```{ path: (the multihash provided to cat), stream: (a readable stream provided for that file) }```

Should I PR a change to the js-ipfs core here https://github.com/ipfs/js-ipfs/blob/master/src/core/ipfs/files.js#L48

to this 

```
const exportStream = Exporter(hash, self._dagS)
exportStream.on('data', (file) => {
      callback(null, file.stream)
})
```

test/files.js
nginnever: I'm confused here, ```stream.on('data', (file) => {``` 
should be labelled  ```stream.on('data', (stream) => {```?

But that is already labelled for the object stream that the core returns. I think you may be wanting the core to be changed like mention <a href="https://github.com/ipfs/interface-ipfs-core/pull/21/files#r64141277">here</a> ?

README.md
diasdavid: yep. The `path` on cat is irrelevant, even misleading

test/files.js
diasdavid: answer here https://github.com/ipfs/interface-ipfs-core/pull/21/files#r64141346

diasdavid:
Is all of that `test-data` required?

test/files.js
```diff
@@ -21,102 +22,94 @@ module.exports = (common) => {
       common.teardown(done)
     })
 
+    describe('.add', () => {
```
nginnever: this fails on js-ipfs-api

```TypeError: Invalid non-string/buffer chunk```

**Comment by nginnever** [comment]
Having trouble writing tests that will work with both js-ipfs core and js-ipfs-api since the requirements vary between the two. 

<a href="https://github.com/ipfs/interface-ipfs-core/pull/20#issuecomment-220788843">Conversation</a> on how to respond from the add function in js-ipfs core here concluded with keeping dagnode objects as the return type for core. 
```
{ path: 'data.txt',
    multihash: <Buffer 12 20 59 94 84 39 06 5f 29 61 9e f4 12 80 cb b9 32 be 52 c5 6d 99 c5 96 6b 65 e0 11 12 39 f0 98 bb ef>,
    size: 4,
    dataSize: 0 
}
``` 
While js-ipfs-api will expect JSON 

```
{ Name: 'data.txt',
    Hash: 'QmVv4Wz46JaZJeH5PMV4LGbRiiMKEmszPYY3g6fjGnVXBS' 
}
```

Cat also can take both buffer and string forms of multihash for input in js-ipfs core, but in js-ipfs-api only string is supported.

```
Error: Invalid ipfs ref path
```

File: README.md [hunk 9]
```diff
@@ -55,6 +55,29 @@ test.all(common)
 
 A valid (read: that follows this interface) IPFS core implementation, must expose the following API.
 
+## Files
+
+### `cat` 
+
+> Streams the data contained by an IPFS object(s) at the given IPFS multihash..
```
diasdavid: You are thinking right, @noffle. Object is a DAGNode, cat returns a File

README.md
```diff
@@ -55,6 +55,29 @@ test.all(common)
...
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.cat(multihash, [callback])
+
+`multihash` is a [multihash][] which can be passed as
+
+- Buffer, the raw Buffer of the multihash (or of and encoded version)
+- String, the toString version of the multihash (or of an encoded version)
```
noffle: I find this wording really unclear -- how can we improve it? Maybe `s/encoded/base58-encoded/`?

Basically what we want to say is: "you can supply a multihash either as a Buffer or a String. Either can be in binary or base58-encoded format."

test/files.js
```diff
@@ -0,0 +1,133 @@
...
+        done()
+      })
+    })
+
+    after((done) => {
+      common.teardown(done)
+    })
+
+    // go-ipfs http and js-ipfs core have different responses
+    describe('.add', () => {
```
noffle: Intentional? I can cover `add` in my PR, rather than duping it here.

README.md
RichardLitt: Well, the grammar is also bad - it should be 'or of an encoded version'. I don't think that encoding has to be base58 (does it?).

nginnever: All Qm strings are currently base58 encoded but they don't have to be. Think the idea is to be able to support more encodings for string represented multihashes in the future. Since base58 encoding the multihash binary to a string is I assume for human readably purposes and if a better binary -> string encoder comes around then others would want it represented that way.

noffle: Yes please: let's be as concise as we can. :tada:

test/files.js
nginnever: promises added :)

test/files.js
nginnever: bad multihash test added

test/files.js
```diff
@@ -0,0 +1,138 @@
...
+          expect(err).to.not.exist
+          file.pipe(bl((err, bldata) => {
+            expect(err).to.not.exist
+            expect(bldata.toString()).to.contain('Check out some of the other files in this directory:')
+            done()
+          }))
+        })
+      })
+
+      // This fails on js-ipfs-api
```
diasdavid: Why? Is there a PR with a fix?

test/files.js
```diff
@@ -0,0 +1,138 @@
...
+            expect(err).to.not.exist
+            expect(bldata.toString()).to.contain('Check out some of the other files in this directory:')
+            done()
+          }))
+        })
+      })
+
+      // You can add a large file to your ipfs repo and change the hash to the file after installing js-ipfs
+      it('returns a large file', (done) => {
+        const hash = 'Qme79tX2bViL26vNjPsF3DP1R9rMKMvnPYJiKTTKPrXJjq'
```
diasdavid: Or we can use .add to get the big file :)

test/files.js
diasdavid: woot :)

test/files.js
nginnever: that's an old comment from before i made this pr to upgrade the js-ipfs-api to accept buffer inputs 

https://github.com/ipfs/js-ipfs-api/pull/280

src/files.js
```diff
@@ -31,169 +32,262 @@ module.exports = (common) => {
...
-        const mh = 'QmVv4Wz46JaZJeH5PMV4LGbRiiMKEmszPYY3g6fjGnVXBS'
-        expect(bs58.encode(res[0].node.multihash()).toString()).to.equal(mh)
-        done()
+        ipfs.files.add(arr, (err, res) => {
+          expect(err).to.not.exist
+          expect(res).to.be.length(1)
+          expect(res[0].path).to.equal('data.txt')
+          expect(res[0].node.size()).to.equal(17)
+          const mh = 'QmVv4Wz46JaZJeH5PMV4LGbRiiMKEmszPYY3g6fjGnVXBS'
+          expect(bs58.encode(res[0].node.multihash()).toString()).to.equal(mh)
```
dignifiedquire: using multihashes would be much nicer

dignifiedquire:
lgtm woul just be nice to replace the b58 calls with multihashes

PR #26, opened 2016-06-05
Interface for files.add

diasdavid:
Builds on top of https://github.com/ipfs/interface-ipfs-core/pull/20

Adds

- [x] Fix aegir setup (rename test->src so that it is buildable
- [x] Stop using require(raw! (the right module was not being loaded, broke all browser tests, we don't need it anyway)
- [x] Stop requiring fs on demand per test, breaks webpack too


package.json
```diff
@@ -2,8 +2,8 @@
   "name": "interface-ipfs-core",
   "version": "0.1.5",
   "description": "A test suite and interface you can use to implement a IPFS core interface.",
-  "main": "test/index.js",
-  "jsnext:main": "test/index.js",
+  "main": "lib/index.js",
```
noffle: Doesn't exist?

noffle:
Thanks for taking this the last mile, @diasdavid.

:horse: LGTM, other than my one comment.

package.json
diasdavid: Get's build on publish, check 

- https://github.com/ipfs/community/blob/master/js-project-guidelines.md
- http://npmjs.org/aegir

PR #32, opened 2016-06-28
block API spec

daviddias:
- block API
  - [x] block.get
  - [x] block.put
  - [x] block.rm
    - [ ] Implement in go-ipfs https://github.com/ipfs/go-ipfs/issues/2914
  - [x] block.stat
- [ ] works in js-ipfs
- [x] works in js-ipfs-api


API/block/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Buffer, the raw Buffer of the multihash 
+- String, the base58 encoded version of the multihash
+
+`callback` must follow `function (err, block) {}` signature, where `err` is an error if the operation was not successful and `block` is a [Block][].
+
+
+```js
+ipfs.block.get(multihash, function (err, data) {
+  // data is the raw data contained in a block
```
diasdavid: of type buffer

diasdavid: Here says that `data` is the raw data, while above it says it is of type `Block`.

API/block/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+
+#### `put`
+
+> Stores input as an IPFS block.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.block.put(data, [callback])
```
diasdavid: `s/data/block`

API/block/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Where `data` can be a
```
diasdavid: s/data/block

API/block/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+- Buffer, requiring that the encoding is specified on the options. if no
```
diasdavid: options are mentioned here, but the function signature doesn't have a field for options.

API/block/README.md
```diff
@@ -0,0 +1,78 @@
...
+`callback` must follow the signature `function (err, stats) {}`, where `err` is
+an error if the operation was not successful and `stats` is an object with
+the format
+
+```JavaScript
+{
+  Key: 'QmPTkMuuL6PD8L2SwTwbcs1NPg14U8mRzerB1ZrrBrkSDD',
+  Size: 10
+}
+```
```
diasdavid: 👍

API/block/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+> Print information of a raw IPFS block.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.block.stat(multihash, [callback])
+
+`multihash` is a [multihash][] which can be passed as:
+
+- Buffer, the raw Buffer of the multihash (or of and encoded version)
```
diasdavid: If it can be a multihash encoded in some format, then a options arg has to exist. See the object API

src/block.js
```diff
@@ -0,0 +1,87 @@
+/* eslint-env mocha */
+/* globals apiClients */
```
diasdavid: why this global?

src/block.js
```diff
@@ -0,0 +1,87 @@
...
+
+const expect = require('chai').expect
+
+module.exports = (common) => {
+  describe.only('.block', () => {
+    const blorbKey = 'QmPv52ekjS75L4JmHpXVeuJ5uX2ecSfSZo88NSyxwA3rAQ'
+    const blorb = Buffer('blorb')
+
+    it('returns an error when putting an array of files', () => {
+      return apiClients.a.block.put([blorb, blorb], (err) => {
```
diasdavid: This seems that it is just prepared for js-ipfs-api and leverages the global.

See https://github.com/ipfs/interface-ipfs-core/blob/master/src/files.js#L26-L35 to learn how to get the `ipfs` object.

src/block.js
```diff
@@ -0,0 +1,87 @@
...
+        console.log(err)
```
diasdavid: probably debugging leftover

src/block.js
```diff
@@ -0,0 +1,87 @@
...
+    })
+
+    it('block.put', (done) => {
+      apiClients.a.block.put(blorb, (err, res) => {
+        expect(err).to.not.exist
+        expect(res).to.have.a.property('Key', 'QmPv52ekjS75L4JmHpXVeuJ5uX2ecSfSZo88NSyxwA3rAQ')
+        done()
+      })
+    })
+
```
diasdavid: Also, follow the same pattern of organisation that other tests have

- .block
  - callback API
    - test 1
    - test 2
    - ...
  - promises API
    - test 1
    - test 2
    - ...

See https://github.com/ipfs/interface-ipfs-core/blob/master/src/files.js for reference

diasdavid:
Going to merge this one before the addition of pull-stream API because we need to do the pull-stream API updates synchronously across the board

PR #34, opened 2016-06-28
pinning API

diasdavid:
> **WIP**

- API calls
  - [x] ipfs.pin.ls
  - [x] ipfs.pin.add
  - [x] ipfs.pin.rm
- [ ] ~~Works in js-ipfs~~ milestone 6
- [x] Works in js-ipfs-api


API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
+pinning API
```
RichardLitt: Pinning

API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pin.add(hash, [options, callback])
+
+Where:
+
+- `hash` is an IPFS multihash.
+- `options` is an object that can contain the following keys:
+  - `recursive` - Recursevely pin the object linked.
```
RichardLitt: recursively

RichardLitt: Default? Type bool?

API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
```
RichardLitt: No res?

API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pin.ls([hash, options, callback])
+
+Where:
+
+- `hash` is an IPFS multihash.
+- `options` is an object that can contain the following keys:
+  - `type` - Return also the type of pin (direct, indirect or recursive)
```
RichardLitt: What about `all`? Should be the default. 

Put these in backticks, they are the strings to pass, not vars.

API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+`callback` must follow `function (err, pinset) {}` signature, where `err` is an error if the operation was not successful. `pinset` is an array of objects with keys `hash` and `type`.
+
+If no `callback` is passed, a promise is returned.
```
RichardLitt: After pinset, or before?

API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Example:
+
+```JavaScript
+ipfs.pin.ls(function (err, pinset) {})
```
RichardLitt: should what the callback would normally do?

API/pinning-api/README.md
RichardLitt: What about `quiet`? 

> Write just hashes of objects. Default: false.

API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+
+#### `rm`
+
+> Remove an hash from the pinset
```
RichardLitt: A hash

RichardLitt: Should be: 
> Removes the pinned object from local storage.

Where is the term pinset coming from?

API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pin.rm(hash, [callback])
+
+Where `hash` is a multihash.
+
```
RichardLitt: What about recursive? Should be a default true bool.

API/pinning-api/README.md
diasdavid: `quiet` is a cli option.

API/pinning-api/README.md
```diff
@@ -0,0 +1,73 @@
...
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
```
RichardLitt: Why no `pinset`? Why no promise?

package.json
```diff
@@ -47,4 +47,4 @@
     "greenkeeperio-bot <support@greenkeeper.io>",
     "nginnever <ginneversource@gmail.com>"
   ]
-}
\ No newline at end of file
+}
```
RichardLitt: Unrelated. Pull out, put in own PR.

API/pinning-api/README.md
diasdavid: The promise is return, but resolved at the same time a callback would be.

API/pinning-api/README.md
diasdavid: Not sure if I understand

src/pin.js
```diff
@@ -0,0 +1,133 @@
...
+
+const testfile = fs.readFileSync(path.join(__dirname, './data/testfile.txt'))
+
+module.exports = (common) => {
+  describe('.pin', () => {
+    let ipfs
+
+    before(function (done) {
+      // CI is slow
+      this.timeout(20 * 1000)
```
RichardLitt: Factor this out, make it a global thing for all tests?

src/pin.js
```diff
@@ -0,0 +1,133 @@
...
+
+      common.setup((err, factory) => {
```
RichardLitt: Where does `common` come from? Curious.

API/pinning-api/README.md
diasdavid: pinset is the list of hashes that are pinned. Only ipfs pin rm and then ipfs repo gc actually removes the object, until gc is performed, the objects will be kept.

diasdavid: We've been using it for a while, it is part of the 'pinning' lingo

src/pin.js
```diff
@@ -0,0 +1,133 @@
...
+    })
+
+    it('add file for testing', (done) => {
+      const expectedMultihash = 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
+
+      ipfs.files.add(testfile, (err, res) => {
+        expect(err).to.not.exist
+
+        expect(res).to.have.length(1)
+        expect(res[0].hash).to.equal(expectedMultihash)
```
RichardLitt: Shouldn't this be a 'pinned <hash>.' or something? Do you just return the hash in js? Isn't that not feature parity?

src/pin.js
```diff
@@ -0,0 +1,133 @@
...
+        expect(res[0].path).to.equal(expectedMultihash)
+        done()
+      })
+    })
+
+    describe('callback API', () => {
+      it('.remove (first, because adding is recursive)', (done) => {
```
RichardLitt: No. Remove should also be recursive.

API/pinning-api/README.md
diasdavid: added a note on the `add` to explain what pinset is :)

src/pin.js
```diff
@@ -0,0 +1,133 @@
...
+        const hash = 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
```
RichardLitt: Make this const declared above, not sure why you declare it so many times.

src/pin.js
```diff
@@ -0,0 +1,133 @@
...
+
+        ipfs.pin.rm(hash, { recursive: true }, (err, res) => {
```
RichardLitt: This isn't an option in the docs. Add it there?

src/pin.js
```diff
@@ -0,0 +1,133 @@
...
+          expect(err).to.not.exist
+          expect(res).to.exist
+          ipfs.pin.ls({ type: 'direct' }, (err, res) => {
```
RichardLitt: Try with all different types?

src/pin.js
```diff
@@ -0,0 +1,133 @@
...
+        return ipfs.pin.add(hash, { recursive: false })
+          .then((res) => {
+            expect(res.Pins[0]).to.be.equal(hash)
+          })
+      })
+
+      it('.ls', () => {
+        return ipfs.pin.ls()
+          .then((res) => {
+            expect(res).to.exist
```
RichardLitt: probably not tight enough.

API/pinning-api/README.md
diasdavid: Promise is return if no callback is passed (says on line below)

src/pin.js
diasdavid: It is just the step of creating the node. If put globally, then we always have to wait 20 seconds for tests to timeout.

API/pinning-api/README.md
RichardLitt: Cool. Yeah, less "What is the new term" and more "Where is it defined". Add it to the glossary? :D

src/pin.js
diasdavid: It is the 'interface test suite pattern', see: https://github.com/ipfs/interface-ipfs-core#usage

API/pinning-api/README.md
RichardLitt: Why not `function (err, Promise) {}` then?

src/pin.js
diasdavid: Since we create an ephemeral node for each batch o test, we have to populate that node.

API/pinning-api/README.md
RichardLitt: How about:

```js
ipfs.pin.ls(function (err, pinset) {
  if (err) return err
  return pinset 
})
```

That's the default callback, no?

src/pin.js
diasdavid: What I meant here was " since ipfs.files.add automatically pins the files, in order to properly test the remaining pin tests, we have to unpin first)"

src/pin.js
RichardLitt: Why not say that, instead of "CI is slow?"

API/pinning-api/README.md
```diff
@@ -0,0 +1,83 @@
...
+- `hash` is an IPFS multihash.
+- `options` is an object that can contain the following keys
+  - `recursive` - Recursively pin the object linked. Type: bool. Default: 'false'
+
+`callback` must follow `function (err, res) {}` signature, where `err` is an error if the operation was not successful. `res` is an array of objects that represent the files that were pinned. Example:
+
+```JavaScript
+{
+  hash: 'QmHash',
+  path: 'some-path
```
RichardLitt: Missing '.

API/pinning-api/README.md
diasdavid: Ah, populate the example, ok :) you got it!

API/pinning-api/README.md
diasdavid: Returning a Promise inside a callback would be an anti-pattern.

src/pin.js
diasdavid: good idea!

src/pin.js
diasdavid: I kind of like of having tests that are really self descriptive, it makes it easier to read and to change.

src/pin.js
diasdavid: thank you for catching that :)

diasdavid:
Thank you @RichardLitt :D Applied almost all your CR, will add a couple more tests to cover remaining cases :)

API/pinning-api/README.md
RichardLitt: Fair enough. Sometimes my mind is not all here.

src/pin.js
RichardLitt: What does the `res` look like, though?

diasdavid: It is an array of objects, as defined in: https://github.com/ipfs/interface-ipfs-core/tree/master/API/files#javascript---ipfsfilesadddata-callback

RichardLitt: Can you test the arr using deepEquals?

API/pinning-api/README.md
```diff
@@ -0,0 +1,91 @@
+Pin API
```
dignifiedquire: The folder is called `pinning-api`, so this probably should be called `Pinning API`

API/pinning-api/README.md
```diff
@@ -0,0 +1,91 @@
...
+If no `callback` is passed, a promise is returned.
+
+Example:
+
+```JavaScript
+ipfs.pin.ls(function (err, pinset) {
+  if (err) {
+    throw err
+  }
+  // console.log(pinset)
```
dignifiedquire: Why is this commented out?

src/pin.js
```diff
@@ -0,0 +1,135 @@
...
+          done()
+        })
+      })
+    })
+
+    after((done) => {
+      common.teardown(done)
+    })
+
+    it('add file for testing', (done) => {
```
dignifiedquire: This shouldn't be a test, it should be part of the `before` block

dignifiedquire: Otherwise if you use `.only` on a test it is not run, resulting in errors.

src/pin.js
```diff
@@ -0,0 +1,135 @@
...
+
+        expect(res).to.have.length(1)
+        expect(res[0].hash).to.equal(expectedMultihash)
+        expect(res[0].path).to.equal(expectedMultihash)
+        done()
+      })
+    })
+
+    describe('callback API', () => {
+      it('.rm (1st, because ipfs.files.add pins automatically)', (done) => {
```
dignifiedquire: the `1st...` should be a comment rather than part of the test description

src/pin.js
```diff
@@ -0,0 +1,135 @@
...
+          expect(err).to.not.exist
+          expect(res.Pins[0]).to.be.equal(hash)
+          done()
+        })
+      })
+
+      it('.list', (done) => {
+        ipfs.pin.ls((err, res) => {
+          expect(err).to.not.exist
+          expect(res).to.exist
```
dignifiedquire: Needs more detailed testing, this could be anything in the result

src/pin.js
```diff
@@ -0,0 +1,135 @@
...
+          })
+        })
+      })
+
+      it('.add', (done) => {
+        const hash = 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
+
+        ipfs.pin.add(hash, { recursive: false }, (err, res) => {
+          expect(err).to.not.exist
+          expect(res.Pins[0]).to.be.equal(hash)
```
dignifiedquire: Can we place lower case these values? , i.e. `.pins`

src/pin.js
```diff
@@ -0,0 +1,135 @@
...
+      it('.rm (1st, because ipfs.files.add pins automatically)', (done) => {
+        const hash = 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
+
+        ipfs.pin.rm(hash, { recursive: true }, (err, res) => {
+          expect(err).to.not.exist
+          expect(res).to.exist
+          ipfs.pin.ls({ type: 'direct' }, (err, res) => {
+            expect(err).to.not.exist
+            expect(res).to.exist
+            expect(res.Keys).to.be.empty
```
dignifiedquire: please lower case `Keys` to `keys`

src/pin.js
```diff
@@ -0,0 +1,135 @@
...
+            done()
+          })
+        })
+      })
+
+      it('.add', (done) => {
+        const hash = 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
+
+        ipfs.pin.add(hash, { recursive: false }, (err, res) => {
```
dignifiedquire: what about `recursive: true`?

src/pin.js
```diff
@@ -0,0 +1,135 @@
...
+          done()
+        })
+      })
+
+      it('.list for a specific hash', (done) => {
+        const hash = 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
+
+        ipfs.pin.ls(hash, (err, res) => {
+          expect(err).to.not.exist
+          expect(res).to.exist
```
dignifiedquire: more detailed testing, that it actually returns the requested hash

src/pin.js
diasdavid: Good point!

diasdavid:
Applied all the CR :)

API/pin/README.md
```diff
@@ -0,0 +1,90 @@
...
+> Remove a hash from the pinset
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pin.rm(hash, [options, callback])
+
+Where:
+- `hash` is a multihash.
+- `options` is an object that can contain the following keys
+  - `recursive` - Recursively unpin the object linked. Type: bool. Default: 'false'
```
dignifiedquire: 'false' -> `false`

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "interface-ipfs-core",
   "version": "0.8.0",
   "description": "A test suite and interface you can use to implement a IPFS core interface.",
-  "main": "lib/index.js",
+  "main": "src/index.js",
```
dignifiedquire: needs removal before merge

src/pin.js
```diff
@@ -0,0 +1,158 @@
...
+        })
+      })
+
+      it('.ls type recursive', (done) => {
+        ipfs.pin.ls({ type: 'recursive' }, (err, pinset) => {
+          expect(err).to.not.exist
+          expect(Object.keys(pinset).length > 0).to.equal(true)
+          done()
+        })
+      })
```
dignifiedquire: These three tests above could be simplified to

function testLs (type) {
  return it(`.ls type ${type}`, (done) => {
  })
}

API/pin/README.md
diasdavid: Richard disagreed on that one, I'll let that 🚲🏠 happen on a separate PR

src/pin.js
diasdavid: Simplified, but not simpler for changes, reading, etc :)

API/pin/README.md
dignifiedquire: @RichardLitt why? (can't find the comment), 'false' in a markdown document looks like it is a string.

src/pin.js
dignifiedquire: fair enough, but please use `.not.empty` rather than `.length > 0).equal(true)`

diasdavid: you got it :)

dignifiedquire:
lgtm

API/pin/README.md
RichardLitt: Yeah, that looks fine to me. This should be default true, though!

PR #36, opened 2016-06-28
DHT API

diasdavid:
> **WIP**
- API calls
  - [x] ipfs.dht.findpeer
  - [x] ipfs.dht.get
  - [x] ipfs.dht.put
  - [x] ipfs.dht.query
  - [x] ipfs.dht.findprovs
- [ ] ~~Works in js-ipfs~~ This will be left for milestone 6 of js-ipfs
- [x] Works in js-ipfs-api
- requires:
  - [x] https://github.com/ipfs/interface-ipfs-core/issues/53


VictorBjelkholm:
Build fails because I've set the `npm test` to fail instead of invalid command. Merging https://github.com/ipfs/interface-ipfs-core/pull/74 will make us a have a valid CI for CircleCI

diasdavid:
@VictorBjelkholm do you have a branch of these tests running on js-ipfs-api?

src/dht.js
```diff
@@ -0,0 +1,62 @@
...
+          ipfs = node
+          done()
+        })
+      })
+    })
+
+    after((done) => {
+      common.teardown(done)
+    })
+    xdescribe('.findpeer', () => {})
```
diasdavid: typo: `xdrescribe`)

diasdavid:
The tests need to be updated to follow the same pattern of the remaining tests in the interface-ipfs-core, the pattern is:

- have a section for "callback API" and heavily test that one
- be creative about tests, if you can make them faster or find bugs in the impl through then, big 👍 
- have a section for "promise API" and test at least each function once

src/dht.js
VictorBjelkholm: Pending test ;)

VictorBjelkholm:
@diasdavid 

> @VictorBjelkholm do you have a branch of these tests running on js-ipfs-api?

Yeah, over here: https://github.com/ipfs/js-ipfs-api/pull/385

> have a section for "callback API" and heavily test that one

:+1: 

> be creative about tests, if you can make them faster or find bugs in the impl through then, big :+1:

:+1: 

> have a section for "promise API" and test at least each function once

Hm, not so sure about this one. The promise API should be tested at one place, to make sure promisify is working correctly, but I don't think we need to test every single function in both the callback way and promise. Makes more sense to have a couple of tests to make sure that one way is working, then always stick with the other otherwise.

diasdavid:
having a test for each method (not all combinations, just a test for each method) has detected in the past missing 'promisify' wraps.

diasdavid:
@VictorBjelkholm I believe you are still on top of this, correct?

VictorBjelkholm:
@diasdavid Yessir!

VictorBjelkholm:
@diasdavid tests are passing, should be ready to be merged so we can move on with https://github.com/ipfs/js-ipfs-api/pull/385

diasdavid:
awesome :D

PR #39, opened 2016-06-30
config API

diasdavid:
- [x] js-ipfs PR https://github.com/ipfs/js-ipfs/pull/340
- [ ] js-ipfs-api PR https://github.com/ipfs/js-ipfs-api/pull/307
  - [ ] needs https://github.com/ipfs/go-ipfs/issues/2927 fixed

### DIFF
--- a/API/config/README.md
+++ b/API/config/README.md
@@ -1,0 +1,52 @@
+config API
+==========
+
+#### `config.get`
+
+> Returns the currently being used config. If the daemon is off, it returns the stored config.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.config.get([key, callback])
+
+`key` is the key of the value that should be fetched from the config file. If no key is passed, then the whole config should be returned. `key` should be of type String.
+
+`callback` must follow `function (err, config) {}` signature, where `err` is an error if the operation was not successful and `config` is a JSON object containing the configuration of the IPFS node.
+
+If no callback is passed, a [promise][] is returned
+
+#### `config.set`
+
+> Adds or replaces a config value.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.config.set(key, value, [callback])
+
+`key` is the key value that will be added or replaced (in case of the value already). `key` should be of type String.
+
+`value` value to be set.
+
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
+
+If no callback is passed, a [promise][] is returned
+
+Note that this operation will **not** spark the restart of any service, i.e: if a config.replace changes the multiaddrs of the Swarm, Swarm will have to be restarted manually for the changes to take difference.
+
+#### `config.replace`
+
+> Adds or replaces a config value.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.config.replace(config, [callback])
+
+`config` is a JSON object that contains the new config.
+
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
+
+If no callback is passed, a [promise][] is returned
+
+Note that this operation will **not** spark the restart of any service, i.e: if a config.replace changes the multiaddrs of the Swarm, Swarm will have to be restarted manually for the changes to take difference.
+
+[promise]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise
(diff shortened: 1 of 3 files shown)


diasdavid:
@xicombd,  @dignifiedquire, @VictorBjelkholm, @nginnever, @noffle, @haadcode, @xicombd would you like to CR this PR and the two related? Namely:

- https://github.com/ipfs/js-ipfs/pull/340
- https://github.com/ipfs/js-ipfs-api/pull/307

Other then the hickup on go-ipfs `config replace` command (https://github.com/ipfs/go-ipfs/issues/2927), it is should be good for the merge :)

src/config.js
```diff
@@ -0,0 +1,179 @@
...
+            expect(err).to.not.exist
+            ipfs.config.get('Fruit', (err, fruit) => {
+              expect(err).to.not.exist
+              expect(fruit).to.equal('morango')
+              done()
+            })
+          })
+        })
+
+        it('set a json object', (done) => {
```
RichardLitt: JSON

API/config/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+#### `config.get`
+
+> Returns the currently being used config. If the daemon is off, it returns the stored config.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.config.get([key, callback])
+
+`key` is the key of the value that should be fetched from the config file. If no key is passed, then the whole config should be returned. `key` should be of type String.
```
dignifiedquire: Can `key` be in the form of `my.config.value`? as far as I know that is available in go

API/config/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+`key` is the key value that will be added or replaced (in case of the value already). `key` should be of type String.
+
+`value` value to be set.
+
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
+
+If no callback is passed, a [promise][] is returned
+
+Note that this operation will spark the restart of any service, i.e: if a config.replace changes the multiaddrs of the Swarm, Swarm will have to be restarted manually for the changes to take difference.
```
dignifiedquire: This is a bit hard to understand, maybe better:

> Note that for this operation to have any effect on running services they have to be restarted manually.

src/config.js
```diff
@@ -0,0 +1,179 @@
...
+        it('retrieve a value through a key', (done) => {
+          ipfs.config.get('Identity', (err, identity) => {
+            expect(err).to.not.exist
+            expect(identity).to.exist
+            done()
+          })
+        })
+
+        it('retrieve a value through a nested key', (done) => {
+          ipfs.config.get('Addresses.Swarm', (err, swarmAddrs) => {
```
dignifiedquire: what about nested arrays?

RichardLitt:
Tried testing; was unable to get the tests to run. The instructions in the Readme weren't clear on how to do this for this module only. 

Otherwise, LGTM.

dignifiedquire:
> Tried testing; was unable to get the tests to run. The instructions in the Readme weren't clear on how to do this for this module only.

The tests can not be run on them own, they need to be integrated into an implementation. See js-ipfs and js-ipfs-api tests for how that looks.

API/config/README.md
nginnever: Agreed, 

>this operation will spark the restart of any service

Makes me think that setting a config value will actually restart the service until reading further where you say you have to manually restart.

API/config/README.md
diasdavid: Not sure if I understand what you mean

API/config/README.md
diasdavid: The sentence is wrong: It should read:

>  Note that this operation will **not** spark the restart of any service

src/config.js
diasdavid: I don't think that is required as in: will ever be used, since a value used by a component of the protocol will always look for a key that belongs to it and not to a partial set of the value of a key.

API/config/README.md
dignifiedquire: There ia no documentation about the fact that you can set nested valuea

src/config.js
dignifiedquire: I disagree. For example it makes sense to only retrieve or set a specific address via the cli, and those are in arrays

diasdavid: And how will you filter those as a User?

API/config/README.md
diasdavid: setting nested values is on `set`

Issue #43, opened 2016-07-06
Create a working example


RichardLitt:
I want to be able to use this. I made a repository following the instructions in this Reamde, I think, and I got nowhere. Can someone help me through this? 

[interface-ipfs-core-example](https://github.com/RichardLitt/interface-ipfs-core-example)

To run:

```
$ mocha test.js
/Users/richard/src/interface-ipfs-core-example/test.js:13
test.all(common)
     ^

TypeError: test.all is not a function
    at Object.<anonymous> (/Users/richard/src/interface-ipfs-core-example/test.js:13:6)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at /Users/richard/.nvm/versions/node/v5.0.0/lib/node_modules/mocha/lib/mocha.js:220:27
    at Array.forEach (native)
    at Mocha.loadFiles (/Users/richard/.nvm/versions/node/v5.0.0/lib/node_modules/mocha/lib/mocha.js:217:14)
    at Mocha.run (/Users/richard/.nvm/versions/node/v5.0.0/lib/node_modules/mocha/lib/mocha.js:469:10)
    at Object.<anonymous> (/Users/richard/.nvm/versions/node/v5.0.0/lib/node_modules/mocha/bin/_mocha:404:18)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Function.Module.runMain (module.js:457:10)
    at startup (node.js:136:18)
    at node.js:972:3
```

dignifiedquire:
Take a look at https://github.com/ipfs/js-ipfs/blob/master/test/core/both/test-files.js

RichardLitt:
Thanks. See #44.

diasdavid:
@RichardLitt hope that @dignifiedquire pointer helped you. This repo are a set of tests that can be used against a working implementation of the ipfs-core api, that's all :)

PR #44, opened 2016-07-06
Updated Usage with working code

RichardLitt:
I was not able to get it working as it was; this should help a bit. See https://github.com/ipfs/interface-ipfs-core/issues/43. Also, updated ToC.


README.md
```diff
@@ -75,22 +60,42 @@ In Go land:
...
+Install `interface-ipfs-core` as one of the dependencies of your project. Install the following test file:
 
-```
-var test = require('interface-ipfs-core')
+```js
+/* eslint-env mocha */
+'use strict'
+
+const test = require('interface-ipfs-core')
+const IPFS = require('ipfs')
```
dignifiedquire: it should be made clear that this does should be your own ipfs instance and not the package `ipfs`.

README.md
```diff
@@ -75,22 +60,42 @@ In Go land:
...
+    })
   },
   teardown: function (cb) {
     cb()
   }
 }
 
-// use all of the test suits
-test.all(common)
+test.files(common)
```
dignifiedquire: This is only an example of running the `files` tests, not all of them, this should be noted with a reference to where one can find a list of all of them

package.json
```diff
@@ -36,7 +36,8 @@
     "readable-stream": "1.1.13"
   },
   "devDependencies": {
-    "aegir": "^3.2.0"
+    "aegir": "^3.2.0",
+    "ipfs": "^0.14.1"
```
dignifiedquire: This is not a good idea, this adds the dependency to all the repos requiring it even though they might not need it

test.js
```diff
@@ -0,0 +1,19 @@
...
+    ipfs.load(() => {
+      cb(null, ipfs)
+    })
+  },
+  teardown: function (cb) {
+    cb()
+  }
+}
+
+test.files(common)
```
dignifiedquire: This should be in something like an `examples` folder or even in another repo, as this is only for show.

README.md
RichardLitt: Where would that be?

test.js
RichardLitt: Moved.

README.md
dignifiedquire: here: https://github.com/ipfs/interface-ipfs-core/blob/master/src/index.js

README.md
```diff
@@ -17,22 +18,6 @@
...
-    - [`object.put`](/API/object#objectput)
-    - [`object.get`](/API/object#objectget)
-    - [`object.data`](/API/object#objectdata)
-    - [`object.links`](/API/object#objectlinks)
-    - [`object.stat`](/API/object#objectstat)
-    - [`object.patch`](/API/object#objectpatch)
-      - [`object.patch.addLink`](/API/object#objectpatchaddlink)
-      - [`object.patch.rmLink`](/API/object#objectpatchrmlink)
-      - [`object.patch.appendData`](/API/object#objectpatchappenddata)
-      - [`object.patch.setData`](/API/object#objectpatchsetdata)
```
dignifiedquire: Why did you remove all of these?

README.md
```diff
@@ -75,22 +60,42 @@ In Go land:
 
 ### Node.js
 
-Install `interface-ipfs-core` as one of the dependencies of your project and as a test file. Then, using `mocha` (for Node.js) or a test runner with compatible API, do:
+Install `interface-ipfs-core` as one of the dependencies of your project. Install the following test file:
 
-```
-var test = require('interface-ipfs-core')
+```js
+/* eslint-env mocha */
```
dignifiedquire: This is not needed, as neither `describe` nor `it` are used in this code snippet.

README.md
```diff
@@ -75,22 +60,42 @@ In Go land:
...
+'use strict'
+
+const test = require('interface-ipfs-core')
+const IPFS = require('ipfs') // This should be your own instance, not the package ipfs.
```
dignifiedquire: Maybe sth like `const IPFS = getYourIPFSInstance()` instead?

README.md
```diff
@@ -75,22 +60,42 @@ In Go land:
...
 
-var common = {
+const common = {
   setup: function (cb) {
-    cb(null, yourIPFSInstance)
+    const ipfs = new IPFS()
+    ipfs.load(() => {
```
dignifiedquire: `ipfs.load` is specific to `js-ipfs` so I am not sure it makes sense to have in here.

README.md
```diff
@@ -75,22 +60,42 @@ In Go land:
...
 
-// use all of the test suits
-test.all(common)
+test.files(common)
+```
+
+Then, using `mocha` (for Node.js) or a test runner with compatible API, do:
+
+```sh
+mocha test.js
```
dignifiedquire: Caan you add `$` in front please, as well as below

examples/test.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-env mocha */
```
dignifiedquire: Why the duplication of this file, with the one in readme?

README.md
diasdavid: or a `>`

README.md
diasdavid: 👍

README.md
RichardLitt: They are not in the file. The Table of Contents should reflect the current file.

README.md
RichardLitt: Can we make this process cleaner? How do I set up my own instance, easily?

README.md
RichardLitt: Should I just remove it?

examples/test.js
RichardLitt: I want to be able to run it, immediately. You can't do that if it is stuck in a README.

README.md
diasdavid: It is the table of contents of the repo, which includes the several APIs spec definitions. Otherwise how will devs be able to open the repo and jump quickly to where they need?

README.md
diasdavid: yes please :)

examples/test.js
diasdavid: There isn't however (or will be) an implementation of IPFS inside this repo.

RichardLitt: There still ought to be a way to run the code, shouldn't there? Otherwise it's not a usable example. See title of the PR.

diasdavid:
@RichardLitt are you familiar with:

- https://github.com/maxogden/abstract-blob-store
- https://github.com/libp2p/interface-transport
- https://github.com/libp2p/interface-connection
- https://github.com/libp2p/interface-stream-muxer
- https://github.com/libp2p/interface-peer-routing
- https://github.com/libp2p/interface-record-store
- https://github.com/libp2p/interface-connection
- https://github.com/libp2p/interface-libp2p

These are all spec definitions plus tests that can be used to make sure that the implementations are compliant.

It would actually be good if https://github.com/ipfs/http-api-spec would have a battery of tests that could be used to test the API, so that we could run them against go-ipfs and js-ipfs. @whyrusleeping how interesting would it be to you to have the CLI extracted from go-ipfs, so that it can be compiled as a standalone binary so that we can put the shardness tests available to run against any IPFS implementation? (Might be better to talk about this in the http-api-spec repo)

RichardLitt:
So it's common to not have runnable usage code? Can we compromise by adding a paragraph explaining out how to set up your own IPFS Instance that can be called programmatically? How would I do that?

RichardLitt:
And yes, please raise an issue about that in http-api-spec. I've been wanting that for a long time, but I can't do it alone.

diasdavid:
https://github.com/ipfs/interface-ipfs-core/pull/44#issuecomment-237845345

Yes, it is common not to have runnable code. Note that these tests are also ran in js-ipfs-api, which doesn't spawn the node at all.

https://github.com/ipfs/interface-ipfs-core/pull/44#issuecomment-237845417

There you go: https://github.com/ipfs/http-api-spec/issues/113

RichardLitt:
Thanks. Ok, changed to yourIPFSInstance, removed example code. 

I still think we need a paragraph on how to spawn your own instance. Can you describe that to me, in simple writing, here?

RichardLitt:
@diasdavid I'd like to figure how how to spawn an instance, so I can write a better doc here about how to do that. Where should I ask that? On IRC?

README.md
```diff
@@ -76,22 +76,38 @@ In Go land:
...
-Install `interface-ipfs-core` as one of the dependencies of your project and as a test file. Then, using `mocha` (for Node.js) or a test runner with compatible API, do:
+Install `interface-ipfs-core` as one of the dependencies of your project. Install the following test file:
 
-```
-var test = require('interface-ipfs-core')
+```js
+'use strict'
+
+const test = require('interface-ipfs-core')
+const yourIPFSInstance = require('ipfs')
```
diasdavid: This doesn't yield an IPFS instance.

diasdavid:
> @diasdavid I'd like to figure how to spawn an instance.

That is the thing, it is not up to the 'spec' to define how to spawn an IPFS instance. The tests only tell you that you have a 'correct' IPFS instance if the thing you pass to it is an IPFS instance that has all the functions and behaviour the tests expect.

RichardLitt:
But you can't run the tests if you don't know how to run an instance. Isn't it important to tell people how to run the tests?

diasdavid:
> But you can't run the tests if you don't know how to run an instance. Isn't it important to tell people how to run the tests?

Not quite, if you want to use this test suite to test 'your own IPFS interface implementation' and you don't know how to create an instance from your implementation, you are in a deeper problem.

Note that this tests get executed in js-ipfs and js-ipfs-api repos, as they are required (which is why the tests are in the src folder, because this is a test suite module). 

Hope this clarifies, closing this PR.

PR #46, opened 2016-08-03
fix: ensure promise tests are correct

dignifiedquire:
This fixes the test failures in ipfs/js-ipfs#392

@diasdavid please review

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -142,30 +142,22 @@ module.exports = (common) => {
 
     describe('promise API', () => {
       describe('.get', () => {
-        it('retrieve the whole config', (done) => {
-          ipfs.config.get()
+        it('retrieve the whole config', () => {
+          return ipfs.config.get()
             .then((config) => {
               expect(config).to.exist
-              done()
-            })
-            .catch((err) => {
-              expect(err).to.not.exist
             })
         })
       })
 
       describe('.set', () => {
-        it('set a new key', (done) => {
-          ipfs.config.set('Fruit', 'banana')
+        it('set a new key', () => {
+          return ipfs.config.set('Fruit', 'banana')
             .then(() => {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "readable-stream": "1.1.13"
   },
   "devDependencies": {
-    "aegir": "^3.2.0"
+    "aegir": "^4.0.0"
```
diasdavid: wanna bump it to 5?

diasdavid:
Could you give more info of what was the failure?

diasdavid:
Have you tested it against js-ipfs-api?

dignifiedquire:
This is the failure: (build: https://circleci.com/gh/ipfs/js-ipfs/665)

> 1) core --both .files promise API .add buffer:
     Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.
    at webpack:///~/interface-ipfs-core/lib/files.js:226:0 <- test/browser.js:162249:13

  

  2) core --both .files promise API .add buffer:
     Resolution method is overspecified. Specify a callback *or* return a Promise; not both.

package.json
dignifiedquire: :D sure

diasdavid:
I'm surprised that this error just happened now. Can you tell what got broken meanwhile?

Otherwise LGTM

dignifiedquire:
The error is part of `mocha@3` which checks for this explicitly, before it just go silently swallowed. The confusing part is that I actually didn't yet upgrade that branch to mocha 3, but it needs fixing in any case.

dignifiedquire:
> Have you tested it against js-ipfs-api?

just tested and things pass

diasdavid:
> mocha@3

Thank you for clarifying :)

Issue #29, opened 2016-06-16
proposal: Files API - dot return DAGNodes on Add


diasdavid:
I've been seeing a lot of users getting confused with the fact that we return DAGNodes on a files add for each file. A DAGNode contains all the information that we might would have hoped for, like the multihash and size of the tree, however, since it has a data field, users are getting confused thinking that the data of that file is on that data field, while in fact, that datafield is just one of the pieces of the whole file.

What if we changed the API to instead of returning DAGNodes, to return just an object with path, multihash and size?

noffle:
That sounds reasonable. +1.

On June 16, 2016 2:21:27 AM PDT, David Dias <notifications@github.com> wrote:
>I've been seeing a lot of users getting confused with the fact that we
>return DAGNodes on a files add for each file. A DAGNode contains all
>the information that we might would have hoped for, like the multihash
>and size of the tree, however, since it has a data field, users are
>getting confused thinking that the data of that file is on that data
>field, while in fact, that datafield is just one of the pieces of the
>whole file.
>
>What if we changed the API to instead of returning DAGNodes, to return
>just an object with path, multihash and size?
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/ipfs/interface-ipfs-core/issues/29

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

VictorBjelkholm:
Yeah, sounds reasonable to me as well. For me, the hardest part is knowing how to get the hash, without knowing everything about the implementation. For example, if I'm adding a file and I want to know the hash of that file, I would have to do the following:

```
res[0].node.toJSON().Hash
```

Whereas I would expect the `Hash` property to sit next to the node `property` or possible inside. Today, even if I log one entry in the `response` array, there is no way to easily see the hash itself, not sure why it's hidden behind the `toJSON` method.

```
{ path: 'a-test-file',
  node:
   DAGNode {
     _cached: null,
     _encoded: null,
     data: <Buffer 08 01>,
     links: [ [Object], [Object], [Object], [Object], [Object], [Object] ] } }
```

diasdavid:
@VictorBjelkholm the 'Hash' is generated through the `multihash()` call, which returns the multihash and then has to be encoded.

[...]

PR #56, opened 2016-08-10
feat(files.add): update files.add API

diasdavid:
@dignifiedquire @haadcode @VictorBjelkholm @noffle @shamb0t @xicombd @jgantunes @nginnever would like your check/:+1: on this :) 

ref: https://github.com/ipfs/interface-ipfs-core/issues/29

- [x] update spec
- [x] update tests
- [x] passes in js-ipfs https://github.com/ipfs/js-ipfs/pull/413
- [x] passes in js-ipfs-api https://github.com/ipfs/js-ipfs-api/pull/345


haadcode:
LGTM!

API/files/README.md
```diff
@@ -28,7 +28,8 @@ If no `content` is passed, then the path is treated as an empty directory
 ```js
 {
   path: '/tmp/myfile.txt',
-  node: DAGNode
+  hash: 'QmHash', // base58 encoded multihash
```
dignifiedquire: This should not be a base58 encoded multihash. Looking at the future we are introducing cid + ipld soon. That means we have to break this interface at that point to a degree. This should be mulithash buffer/object something like that.

diasdavid: Thank you, I was thinking of that too, specially with the introduction of CID.

However, there will be a CID string version anyway, so the question emerges, do we have a raw multihash/CID or the string version

dignifiedquire: CIDs will allow for different versions of encoding, i.e. non base58. This interface though binds itself to using base58 which I find the most concerning part.

diasdavid: got it. Right now, the content Ids are presented as base58 Encoded multihash, because there is no multibase nor multicodec. Once we have CID, the description will change to 'String representation of the CID', cause it is a self described Id :D

diasdavid:
I'm finishing up the changes in js-ipfs-api. Please review these meanwhile if you can.

diasdavid:
Done both, review please :)

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "interface-ipfs-core",
   "version": "0.7.2",
   "description": "A test suite and interface you can use to implement a IPFS core interface.",
-  "main": "lib/index.js",
+  "main": "src/index.js",
```
dignifiedquire: change before merge :P

src/files.js
```diff
@@ -53,21 +53,27 @@ module.exports = (common) => {
...
             expect(err).to.not.exist
             expect(res).to.be.length(1)
-            expect(res[0].path).to.equal('data.txt')
-            expect(res[0].node.size()).to.equal(17)
-            const mh = 'QmVv4Wz46JaZJeH5PMV4LGbRiiMKEmszPYY3g6fjGnVXBS'
-            expect(bs58.encode(res[0].node.multihash()).toString()).to.equal(mh)
+            const file = res[0]
+            expect(file.path).to.equal('data.txt')
+            expect(file.size).to.equal(17)
+            expect(file.hash).to.equal(expectedMultihash)
```
dignifiedquire: Might be nicer to write this like so

```js
expect(file).to.be.eql({
  path: 'data.txt',
  size: 17,
  hash: expectedMultihash
})
```

or 

```js
expect(file).to.have.key('path', 'data.txt')
expect(file).to.have.key('size', 17)
expect(file).to.have.key('hash', expectedMultihash)
```

dignifiedquire:
except the two things above all lgtm

src/files.js
diasdavid: Changed the first case to the first one, then I changed the remaining ones to the second options, but those got angry at me:

```
      Uncaught AssertionError: expected { Object (path, hash, ...) } to have key 'QmVvjDy7yF7hdnqE8Hrf4MHo5ABDtb5AbX6hWbD3Y42bXP'
      + expected - actual

       [
      -  "hash"
      -  "path"
      -  "size"
      +  "QmVvjDy7yF7hdnqE8Hrf4MHo5ABDtb5AbX6hWbD3Y42bXP"
       ]

```

It seems that .to.have.key expects that the arguments are the keys to be looked inside the object and not their values. Checked the Chai API docs and didn't find an example to check for a key and then assert its value (http://chaijs.com/api/bdd/), keeping those how they were :)

diasdavid:
Thank you for the CR @dignifiedquire :):)

src/files.js
dignifiedquire: ahh sorry, it should have been `.to.have.property('key', 'value')`

PR #57, opened 2016-08-10
Use ipfs-factory to launch nodes

diasdavid:
- [x] works in js-ipfs
- [x] works in js-ipfs-api


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "interface-ipfs-core",
   "version": "0.6.0",
   "description": "A test suite and interface you can use to implement a IPFS core interface.",
-  "main": "lib/index.js",
+  "main": "src/index.js",
```
dignifiedquire: remove before merge

diasdavid: ;)

PR #71, opened 2016-09-09
Check files.get includes paths/content instead of deepEqual

victorb:
This ensures that the files and their path is in the response and ignores the order that we cannot trust


VictorBjelkholm:
Needed for https://github.com/ipfs/js-ipfs/pull/323

PR #72, opened 2016-09-14
refactor(files): call cat from files.cat like the interface defines it

diasdavid:


diasdavid:
This is a minor change to make it conform with the spec, but that enables js-ipfs to stop having to alias .cat (which was the only method being aliased for the sake of these tests). tested it in js-ipfs and js-ipfs-api

PR #74, opened 2016-09-17
Run lint instead of test in CircleCI

VictorBjelkholm:
Instead of running npm test that should fail, lets just print that we have no tests and then exit cleanly

Fixes failing build from here: https://github.com/ipfs/interface-ipfs-core/pull/36


diasdavid:
Seems circle is not happy due to a missing dependency, I've give it a kick to restart with no cache, let's see if that was it.

VictorBjelkholm:
@diasdavid passing now :+1:

PR #95, opened 2016-11-23
feat: update to awesome dag-pb

diasdavid:
> WIP

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -36,11 +36,9 @@ module.exports = (common) => {
         it('no layout', (done) => {
           ipfs.object.new((err, node) => {
             expect(err).to.not.exist
-            node.toJSON((err, obj) => {
-              expect(err).to.not.exist
-              expect(obj.Hash).to.equal('QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n')
-              done()
-            })
+            const nodeJSON = node.toJSON()
+            expect(nodeJSON.multihash).to.equal('QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n')
+            done()
           })
         })
       })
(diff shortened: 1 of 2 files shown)


src/object.js
```diff
@@ -222,59 +231,65 @@ module.exports = (common) => {
...
-          let node1Retrieved
+          let node1a
+          let node1b
+          let node1c
+          let node2
 
           series([
             (cb) => {
-              node1.addNodeLink('some-link', node2, cb)
+              DAGNode(new Buffer('Some data 1'), (err, node) => {
```
haadcode: Should this be DAGNode.create()?

src/object.js
```diff
@@ -222,59 +231,65 @@ module.exports = (common) => {
...
+                expect(err).to.not.exist
+                node1a = node
+                cb()
+              })
             },
             (cb) => {
-              ipfs.object.put(node1, (err, node) => {
+              DAGNode(new Buffer('Some data 2'), (err, node) => {
```
haadcode: Should this be DAGNode.create()?

haadcode:
Couple of questions. Otherwise looks good to me 👍

src/object.js
```diff
@@ -163,22 +158,44 @@ module.exports = (common) => {
...
+                cb()
+              })
+            },
+            (cb) => {
+              DAGNode.create(new Buffer('Some data 2'), (err, node) => {
+                expect(err).to.not.exist
+                node2 = node
+                cb()
+              })
+            },
```
dignifiedquire: Same as in https://github.com/ipld/js-ipld-resolver/pull/67#discussion_r89480770

src/object.js
```diff
@@ -469,38 +451,56 @@ module.exports = (common) => {
...
+            (cb) => {
+              DAGNode.create(new Buffer('Some data 2'), (err, node) => {
                 expect(err).to.not.exist
+                node2 = node
+                cb()
+              })
+            },
```
dignifiedquire: Same as in https://github.com/ipld/js-ipld-resolver/pull/67#discussion_r89480770

src/object.js
```diff
@@ -554,52 +547,71 @@ module.exports = (common) => {
...
               })
-            })
-          })
+            },
+            (cb) => {
+              DAGNode.create(new Buffer('Some data 2'), (err, node) => {
+                expect(err).to.not.exist
+                node2 = node
+                cb()
+              })
```
dignifiedquire: Same as in Same as in https://github.com/ipld/js-ipld-resolver/pull/67#discussion_r89480770

src/object.js
diasdavid: thank you, just fixed now with more testing :)

src/object.js
diasdavid: absolutely correct, missed 2

diasdavid:
Thank you for the review :)

src/object.js
diasdavid: True, it can save us some lines, however, I don't dislike how it is right now, it makes it very easy to read (I believe). Just for the sake of time (and wanting to have this before the hackathon), will leave this reconsideration for after.

dignifiedquire: I find it super hard to read :/ as it is spread on so many lines that I have difficulty figuring out what is actually happening

PR #101, opened 2016-12-05
feature: add pubsub (tests and API Spec)

haadcode:
This PR will add tests for the upcoming pubsub API.

Note that this PR is dependent on https://github.com/ipfs/js-ipfs-api/pull/465.


src/pubsub-message.js
```diff
@@ -0,0 +1,82 @@
+/* eslint-env mocha */
+'use strict'
+
+const expect = require('chai').expect
+const isNode = require('detect-node')
+const PubsubMessage = require('../../src/pubsub-message') // eslint-disable-line no-unused-vars
```
haadcode: We need it eventually. Leaving it there as a placeholder.

However, these should not be required from '../src' but from ipfs-api. @dignifiedquire how can I do that? interface-ipfs-api tests that refer to components in js-ipfs-api (and the same components will prolly be used in js-ipfs) but are not publicly exposed by the module?

src/pubsub.js
```diff
@@ -0,0 +1,508 @@
...
+              process.nextTick(() => loop())
+            })
+          } else {
+            done()
+          }
+        }
+        loop()
+      })
+
+      it('call subscribe 1k times', (done) => {
```
VictorBjelkholm: This is implicitly tested in the subscribe/unsubscribe 1k times, so we can remove this test.

src/pubsub.js
```diff
@@ -0,0 +1,508 @@
...
+
+          waitForPeers(ipfs, [ipfs2.PeerId], (err) => {
+            expect(err).to.not.exist
+            startTime = new Date().getTime()
+            loop()
+          })
+        })
+      })
+
+      it('call publish 1k times', (done) => {
```
VictorBjelkholm: This is implicitly tested in the subscribe/unsubscribe 1k times, so we can remove this test.

src/pubsub.js
haadcode: Unsubscribe changes the internal state. This test is meant to make sure that NOT calling unsubscribe before calling subscribe again doesn't change the state and works as expected.

src/pubsub.js
haadcode: Same as above.

src/pubsub-message.js
```diff
@@ -0,0 +1,82 @@
...
+      expect(message.topicIDs[0]).to.equal(topicName)
+    } catch (e) {
+      expect(e).to.not.exist
+    }
+  })
+
+  describe('immutable properties', () => {
+    const message = PubsubMessageUtils.create('A', 'hello', '123', ['hello world'])
+
+    it('from', () => {
```
haadcode: Fixed.

src/pubsub-message.js
dignifiedquire: you shouldn't require them, the should be passed in through the `setup` function

src/pubsub-message.js
VictorBjelkholm: Not sure if that's the right fix. You just added a expect in the try part, which would not get run if there is an error and now we're doing different assertions depending on if the code throws an exception or not. The tests should not be able to change the executed code path.

src/pubsub-message.js
haadcode: How? How passed what? Can you provide an example?

dignifiedquire: The tests have a `before` filter, which calls `common.setup`, so you can pass more things through there. You could either add the functionality to the `factory` or as a third parameter.

src/pubsub-message.js
haadcode: That's the expected behaviour we're testing here: does it fire an error? The expect I added to the try doesn't get run if an error occurs so we're good. 

I changed it a bit and just pushed a new commit to make it even more implicit that we're testing immutability here: check that the values are the same as before if no error is thrown.

src/pubsub.js
```diff
@@ -0,0 +1,586 @@
...
+        })
+
+        // TODO clarify what is the goal of pubsub.peers
+        it('returns an error when not subscribed to a topic', (done) => {
+          ipfs1.pubsub.peers(topic, (err, peers) => {
+            expect(err).to.exist
+            expect(err.toString()).to.equal(`Error: Not subscribed to '${topic}'`)
+            done()
+          })
+        })
```
diasdavid: @haadcode why would it fail if I'm not subscribed to a topic? Could you clarify?

haadcode: My logic here was that the state of the subscriptions is the source of truth: if you're subscribed, you can query it, cancel it, etc. If you're not, you can't.

What would you expect to receive if you're not subscribed to the topic and call .peers()? En empty list?

Say we have a use case like in the tests for the "waitForPeers", we callback only after we received peers. We don't subscribe to the topic but we keep polling pubsub.peers in the hopes that it returns the peer some day. We keep waiting and waiting and it doesn't seem to work, we check connectivity, we start looking elsewhere, because peers just returns an empty array. That's pretty confusing behaviour for the API. In this case, I believe it's better to tell the user with an error that "hey, what' you're maybe trying to do, will not work".

haadcode:
I'm ok merging this when you are.

src/pubsub.js
diasdavid: Well, for me `peers(topic)` tells me I should expect "give me the peers I have in my peerSet that are subscribed to a given topic".

src/pubsub.js
```diff
@@ -143,7 +154,7 @@ module.exports = (common) => {
           })
         })
 
-        it.skip('returns no peers within 10 seconds', (done) => {
+        it('returns no peers within 10 seconds', (done) => {
```
diasdavid: @haadcode this test is failing in `js-ipfs-api`

src/pubsub.js
```diff
@@ -159,7 +170,7 @@ module.exports = (common) => {
           })
         })
 
-        it.skip('doesn\'t return extra peers', (done) => {
+        it('doesn\'t return extra peers', (done) => {
```
diasdavid: @haadcode this one too

src/pubsub.js
```diff
@@ -182,10 +193,16 @@ module.exports = (common) => {
           })
         })
 
-        it.skip('returns peers for a topic - one peer', (done) => {
+        it('returns peers for a topic - one peer', (done) => {
```
diasdavid: @haadcode and this one

src/pubsub.js
```diff
@@ -182,10 +193,16 @@ module.exports = (common) => {
...
           // Currently go-ipfs returns peers that have not been subscribed to the topic
           // Enable when go-ipfs has been fixed
```
diasdavid: Why not filter at js-ipfs-api?

diasdavid:
> I'm ok merging this when you are.

Found some issues when running the new tests you added/enabled in js-ipfs-api. They run well in js-ipfs though.

haadcode:
> Found some issues when running the new tests you added/enabled in js-ipfs-api. They run well in js-ipfs though.

@diasdavid this is due to using a (old) version go-ipfs that has bugs for those tests. As far as I know they're now fixed in master but we need to wait for go-ipfs 0.4.5-rc1 release to verify the tests are working.

Shall we put this to "blocked" until we have that?

src/pubsub.js
```diff
@@ -0,0 +1,575 @@
...
+                }
+              })
+
+              function loop () {
+                if (sendCount < count) {
+                  sendCount++
+                  outputProgress()
+                  ipfs2.pubsub.publish(topic, expectedString, (err) => {
+                    expect(err).to.not.exist
+                    process.nextTick(() => loop())
```
diasdavid: @haadcode this is till Node.js specific

API/pubsub/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pubsub.subscribe(topic, options, callback)
+
+- `topic` - type: String
+- `options` - type: Object, optional, might contain the following properties:
+  - `discover`: type: Boolean - Will use the DHT to find 
+
+`callback` must follow `function (err, subscription) {}` where Subscription is a Node.js Stream in Object mode, emiting a `data` event for each new message on the subscribed topic.`err` is an error if the operation was not successful.
```
dignifiedquire: I don't think this is a good interface. We all know the issues of node.js streams and using object streams is even more problematic, as some modules (like hapi.js) don't even work with object streams.

I would propose a simpler api using raw `EventEmitter`.

```js
const handler = onMessage
ipfs.pubsub.subscribe('mytopic', {discover: true}, onMessage)
ipfs.pubsub.unsubscribe('mytopic', onMessage)
```

Which would be implemented using `EventEmitter` and mostly aliases for `.on` and `removeListener`. Where `removeListener` does a check onto [emitter.listenerCount(topic)](https://nodejs.org/docs/latest/api/events.html#events_emitter_listenercount_eventname) and if that is `0` it will actually unsubscribe from the underlying thing.

diasdavid: 💯

API/pubsub/README.md
```diff
@@ -0,0 +1,63 @@
...
+> Returns the list of subscriptions the peer is subscribed to.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pubsub.ls(topic, callback)
+
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
+
+If no `callback` is passed, a [promise][] is returned.
+
```
dignifiedquire: Missing docs on the value that is resolved

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "interface-ipfs-core",
-  "version": "0.22.0",
+  "version": "0.22.1",
```
dignifiedquire: why is there a version bump in a PR?

src/object.js
```diff
@@ -41,6 +41,19 @@ module.exports = (common) => {
             done()
           })
         })
+
+        it.skip('template unixfs-dir', (done) => {
```
dignifiedquire: this seems entirely unrelated

src/pubsub.js
```diff
@@ -0,0 +1,561 @@
...
+                subscription.cancel(done)
+              })
+            })
+          })
+
+          it('list with 3 subscribed topicss', (done) => {
+            const topics = ['one', 'two', 'three']
+            series(
+              topics.map((t) => (cb) => ipfs1.pubsub.subscribe(t, cb))
+            , (err, subs) => {
```
dignifiedquire: please put the comma at the end of the line

package.json
diasdavid: same as https://github.com/ipfs/interface-ipfs-core/pull/101#discussion_r92939110

API/pubsub/README.md
diasdavid: This will require changes in the js-ipfs-api PR too

src/pubsub.js
```diff
@@ -156,7 +157,8 @@ module.exports = (common) => {
             })
           })
 
-          it('returns no peers within 10 seconds', (done) => {
+          // I don't understand the purpose of this test
+          it.skip('returns no peers within 10 seconds', (done) => {
```
diasdavid: @haadcode ^^

API/pubsub/README.md
haadcode: Absolutely. Let's do it!

diasdavid: Awesome! Please reuse open connections in js-ipfs-api, i.e: If a subscription request is already open, avoid opening a new one, just increase the ref count.

gavinmcdermott: @dignifiedquire - Awesome. Let me know if there's any areas you need help / support in this!

src/pubsub.js
```diff
@@ -0,0 +1,569 @@
...
+        before((done) => {
+          parallel([
+            (cb) => ipfs1.swarm.connect(ipfs2.peerId.addresses[0], cb),
+            (cb) => ipfs2.swarm.connect(ipfs3.peerId.addresses[0], cb),
+            (cb) => ipfs1.swarm.connect(ipfs3.peerId.addresses[0], cb)
+          ], done)
+        })
+
+        describe('.peers', () => {
+          it('returns an error when not subscribed to a topic', (done) => {
```
diasdavid: There was an open question in this interface. I personally don't understand why the question "Tell me who I know that is subscribed to a topic" needs to fail in case I'm not interested to the topic.

@haadcode could you clarify your argument for it to be this way? 

@gavinmcdermott @dignifiedquire thoughts?

src/pubsub.js
```diff
@@ -0,0 +1,569 @@
...
+                new RegExp(`Not subscribed to '${topic}'`)
+              )
+
+              done()
+            })
+          })
+
+          it.skip("doesn't return extra peers", (done) => {
+            // Currently go-ipfs returns peers that have not been
+            // subscribed to the topic. Enable when go-ipfs has been fixed
```
diasdavid: What this means is that `js-ipfs-api` should filter out those who aren't subscribed

src/pubsub.js
```diff
@@ -0,0 +1,569 @@
...
+
+            const topicOther = topic + 'different topic'
+
+            series([
+              (cb) => ipfs1.pubsub.subscribe(topic, sub1, cb),
+              (cb) => ipfs2.pubsub.subscribe(topicOther, sub2, cb)
+            ], (err) => {
+              expect(err).to.not.exist
+              setTimeout(() => {
+                ipfs1.pubsub.peers(topic, (err, peers) => {
```
diasdavid: Are these `peers`, peerInfo objects?

src/pubsub.js
```diff
@@ -0,0 +1,569 @@
...
+              expect(inbox1.sort()).to.be.eql(outbox.sort())
+              expect(inbox2.sort()).to.be.eql(outbox.sort())
+
+              done(err)
+            })
+
+            function sub1 (msg) {
+              inbox1.push(msg.data.toString())
+              // TODO: enable when go-ipfs is unbroken
+              // expect(msg.from).to.be.eql(ipfs2.peerId.id)
```
diasdavid: Is there a issue that can also be linked here?

API/pubsub/README.md
```diff
@@ -0,0 +1,63 @@
...
+##### `JavaScript` - ipfs.pubsub.publish(topic, data, callback)
+
+- `topic` - type: String
+- `data` - type: Buffer
+
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
+
+If no `callback` is passed, a [promise][] is returned.
+
```
diasdavid: Please complete

API/pubsub/README.md
```diff
@@ -0,0 +1,63 @@
...
+> Returns the peers that are subscribed to one topic.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pubsub.peers(topic, callback)
+
+- `topic` - type: String
+
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
```
diasdavid: Please complete

src/pubsub.js
haadcode: I'm absolutely fine not throwing an error.

dignifiedquire: Alright, will change to not throwing behaviour

src/pubsub.js
dignifiedquire: they are peer ids encoded as b58 multihash

dignifiedquire: so strings

src/pubsub.js
dignifiedquire: yes, it's mentioned in the first case, I just didn't want to duplicated it all over the file

API/pubsub/README.md
dignifiedquire: done

dignifiedquire:
Ready for next review

API/pubsub/README.md
```diff
@@ -0,0 +1,72 @@
...
+#### `pubsub.publish`
+
+> Publish a data message to a pubsub topic.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pubsub.publish(topic, data, callback)
+
+- `topic: string`
+- `callback: (Error) => ()` - Calls back with an error or nothing if the publish was successfull.
```
diasdavid: It is missing the description of `data`

src/pubsub.js
```diff
@@ -0,0 +1,573 @@
...
+
+      after((done) => {
+        common.teardown(done)
+      })
+
+      describe('single node', () => {
+        describe('.publish', () => {
+          it('message from string', (done) => {
+            ipfs1.pubsub.publish(topic, 'hello friend', done)
+          })
```
diasdavid: I'm worried with accepting plain strings as argument. The issue that we had last week Monday was because the initial pubsub code @haadcode wrote would always convert any value to a Buffer on the publish implementation, but then, because the tests were written like `expect(dataReceived).to.equal('some string')`, the implementation had a `data.toString()` internally as a quick patch, and when I changed that to make it an operation in the outside (like we have here now, see below, preserving the buffer type), it made orbit stop working because there was an assumption that 'data' was a string, since orbit publishes strings (or was publishing).

Strings and Buffers are one of banes of JavaScript, it would be better to force it to be just Buffer and no other type (creating a test here to make sure it fails with anything non buffer), to avoid future problems.

src/pubsub.js
diasdavid: I see, fair point :( Is there a feature request for the go-ipfs implementation? @whyrusleeping @Kubuxu ?

src/pubsub.js
diasdavid: 👍

src/pubsub.js
Kubuxu: I think it was resolved recently, can you give it another try?

src/pubsub.js
dignifiedquire: We could also check strictly for string or Buffer, but nothing else, which makes it a bit nicer on the user, but avoids issues like passing an array by accident.

src/pubsub.js
diasdavid: It would be awesome to have a way to fetch peer info from the http api, namely known multiaddrs, public key, supported transports, etc //cc @whyrusleeping

API/pubsub/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+> Subscribe to a pubsub topic.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pubsub.subscribe(topic, options, handler, callback)
+
+- `topic: string`
+- `options: Object` - (Optional), might contain the following properties:
+  - `discover`: type: Boolean - Will use the DHT to find
```
diasdavid: "to find.." - sentence needs to be finished

API/pubsub/README.md
diasdavid: 🛎

API/pubsub/README.md
```diff
@@ -0,0 +1,72 @@
...
+> Publish a data message to a pubsub topic.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.pubsub.publish(topic, data, callback)
+
+- `topic: string`
+- `callback: (Error) => ()` - Calls back with an error or nothing if the publish was successfull.
+
+- Returns: `Promise` if no `callback` was passed otherwise `undefined`.
```
diasdavid: We have used always: `If no `callback` is passed, a [promise][] is returned.`, I'm fine changing the sentence, but I do like consistency :)

src/object.js
```diff
@@ -42,7 +42,7 @@ module.exports = (common) => {
           })
         })
 
-        it('template unixfs-dir', (done) => {
+        it.skip('template unixfs-dir', (done) => {
```
diasdavid: Why do we need to skip this one?

src/pubsub.js
```diff
@@ -0,0 +1,573 @@
...
+  return (err) => {
+    if (err) {
+      return done(err)
+    }
+
+    if (++i === n) {
+      done()
+    }
+  }
+}
```
diasdavid: Can't we use the `chai` checkmark thing?

src/pubsub.js
diasdavid: * needs to be resolved before the merge

src/pubsub.js
```diff
@@ -0,0 +1,573 @@
...
+                // Now all listeners are gone no subscription anymore
+                expect(res[2]).to.be.eql([])
+                check()
+              })
+            }
+
+            const handler2 = (msg) => {
+              expect(msg.data.toString()).to.be.eql('hello')
+              check()
+            }
```
diasdavid: We've been kind of declaring functions always with the word `function`, it has some benefits as making it easy to grep. We kind of have followed (naturally) the pattern of always using `function` throughout the code and I would like to keep it consistent, unless there is a strong case not to.

src/pubsub.js
```diff
@@ -0,0 +1,573 @@
...
+          parallel([
+            (cb) => ipfs1.swarm.connect(ipfs2.peerId.addresses[0], cb),
+            (cb) => ipfs2.swarm.connect(ipfs3.peerId.addresses[0], cb),
+            (cb) => ipfs1.swarm.connect(ipfs3.peerId.addresses[0], cb)
+          ], (err) => {
+            if (err) {
+              return done(err)
+            }
+            // give some time to let everything connect
+            setTimeout(done, 300)
```
diasdavid: Does this need to be here? Will it fail it it doesn't?

src/pubsub.js
haadcode: I agree @diasdavid that this is something we should decide now as it has bitten us already in a bad way. What do we do in other function that accept similar data a param, like object.put? It might be a good idea to follow similar convention. If it's string+buffer, then what David proposes makes sense (always convert to buffer and make it as explicit to the user as possible (docs + erroring on anything but string or buffer). If it's only a buffer, it would also make sense to me (input buffer, receive buffer). This is how I tried to approach it in js-ipfs-api: everything we receive is a string, so inputs should be strings (from user's perspective). Obviously we would need to change js-ipfs-api to convert all strings to buffers, and probably in a lot of use cases the user then converts the buffer back to a string, but I suppose it's a cost we could live with (and on a positive side, js-ipfs would be tiny bit faster because it only needs to use buffers).

So... I think I like the idea of accepting buffers-only the most, but it wouldn't be too bad if we accepted strings too.

dignifiedquire: alright, going to change things to buffer only then

API/pubsub/README.md
dignifiedquire: I think this is a great sentence :P

src/object.js
dignifiedquire: I don't even know why this is here in the first place

src/pubsub.js
dignifiedquire: I honestly found this to be simpler to use and understand. But not opposed to switching.

src/pubsub.js
dignifiedquire: You have used `function` I try to use `const name = ()` especially inside other functions and to avoid using functions before they are declared, as it is very unfortunate that `function` declarations are hoisted to the top level, allowing for code to use functions before they have been declared.

src/pubsub.js
dignifiedquire: yes

dignifiedquire: sometimes

src/object.js
dignifiedquire: fixed

API/pubsub/README.md
dignifiedquire: changed back

src/pubsub.js
dignifiedquire: done

API/pubsub/README.md
dignifiedquire: fixed

API/pubsub/README.md
```diff
@@ -3,38 +3,47 @@ pubsub API
...
 
-> Publish a data message to a pubsub topic
+> Publish a data message to a pubsub topic.
 
 ##### `Go` **WIP**
 
 ##### `JavaScript` - ipfs.pubsub.publish(topic, data, callback)
 
-- `topic` - type: String
-- `data` - type: Buffer
```
dignifiedquire: fixed

src/pubsub.js
diasdavid: Sounds good :)

dignifiedquire:
🎉  100 comments

PR #105, opened 2017-01-09
fix/identify finish

diasdavid:
Added a short time delay after connect, because now with identify and pubsub handshake, tests were finishing before than those handshakes ended, making the streams close before multistream finished.


Kubuxu:
Is that standard practice in js? In Go we try not to use timings for anything in tests as they are generally unreliable, due to possible CI hickups, load spikes, swap and so on, and thus can cause random failures.

diasdavid:
Definitely not a good practice, the alternative is to make some operations lock, which will make things slower for running nodes, but always passing tests in CI.

What we see here is that we drop the listeners so fast that the pubsub multistream handshake doesn't have time to finish.

diasdavid:
actually, nvm, thanks @Kubuxu <3

PR #106, opened 2017-01-10
WIP: support the new unixfs-engine balanced builder

pgte:
Changes in tests reflecting the new results from the new default balanced builder in unixfs-engine.

Depends on merging https://github.com/ipfs/js-ipfs-unixfs-engine/pull/118

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -52,7 +52,7 @@ module.exports = (common) => {
       describe('.add', () => {
         it('stream', (done) => {
           const buffered = new Buffer('some data')
-          const expectedMultihash = 'QmVv4Wz46JaZJeH5PMV4LGbRiiMKEmszPYY3g6fjGnVXBS'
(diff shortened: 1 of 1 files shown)


diasdavid:
I've run these changes with `js-ipfs-api` against `go-ipfs` and since the expected hashes changed, the tests fail, they both should yield the same hashes for the same files

I checked to see if it could have been the change from raw to file (that could had happen in the other direction), but it seems that unixfs-engine is good with regards to that.

pgte:
Nevermind, I hadn't realised that the test files were relatively small.
By the way, the difference with go-ipfs is that, for a file that is only one chunk, go-ipfs does not put a single link parent in front of it (going against what it does when recursing).
Tests against https://github.com/ipfs/js-ipfs-unixfs-engine/pull/118 should be passing now.

PR #109, opened 2017-01-12
tests: make sure load tests clean after themselves

diasdavid:


src/pubsub.js
```diff
@@ -55,7 +55,7 @@ function makeCheck (n, done) {
 }
 
 module.exports = (common) => {
-  describe('.pubsub', () => {
+  describe.only('.pubsub', () => {
```
dignifiedquire: please remove :)

diasdavid: right!

PR #110, opened 2017-01-17
fix(config): make promise tests actually test the full promise chain

diasdavid:
all we need for `js-ipfs 0.21`:)

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -158,11 +158,9 @@ module.exports = (common) => {
       describe('.set', () => {
         it('set a new key', () => {
           return ipfs.config.set('Fruit', 'banana')
-            .then(() => {
-              ipfs.config.get('Fruit', (err, fruit) => {
-                expect(err).to.not.exist
-                expect(fruit).to.equal('banana')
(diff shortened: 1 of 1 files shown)


src/config.js
```diff
@@ -158,12 +158,11 @@ module.exports = (common) => {
...
-            .then(() => {
-              ipfs.config.get('Fruit', (err, fruit) => {
-                expect(err).to.not.exist
-                expect(fruit).to.equal('banana')
-              })
+            .then(() => ipfs.config.get('Fruit'))
+            .then((fruit) => {
+              expect(fruit).to.equal('banana')
             })
+            .catch((err) => expect(err).to.not.exist)
```
VictorBjelkholm: Shouldn't need the catch here since we're returning the promise so mocha should automatically catch any errors.

diasdavid: right

diasdavid: fixed :)

diasdavid:
Thank you for the CR @VictorBjelkholm :)

PR #111, opened 2017-01-17
test: last assertions are fine in js-ipfs

diasdavid:
no reason not to have these assertions as part of the suite, tested in js-ipfs, they run fine


dignifiedquire:
Yeah they were only disabled because they were failing in js-ipfs-api, due to the mentioned go-ipfs encoding issues.

PR #112, opened 2017-01-31
feat: DAG API basics (get, put, rm)

diasdavid:
- [x] implementation js-ipfs
- [ ] implementation in js-ipfs-api
  - [ ] https://github.com/ipfs/js-ipfs-api/pull/520


src/dag.js
```diff
@@ -0,0 +1,141 @@
...
+          pbNode = node
+          done()
+        })
+
+        cborNode = {
+          data: someData
+        }
+      })
+
+      describe('.put', () => {
```
dignifiedquire: You should also get these nodes to check that they have been actually put in the correct way

diasdavid: that is what happens in the get tests, this way put is tested in isolation and in conjunction with get

diasdavid:
I'm divided in waiting for feedback on DAG API things and moving forward, I've been kind of waiting for more throughout feedback since last quarter and now I've so many 'little' things in my plate that it is impossible to achieve something during two weeks if I wait for a lot of RTT

I'll just move fast for the dag.get dag.put and dag.resolve API in js-ipfs land and bring up what I need to happen in go-ipfs land to make it go through js-ipfs-api

@whyrusleeping @nicola

src/dag.js
dignifiedquire: But not for all the formats as far as I can see.

PR #114, opened 2017-01-31
feat: dag.workbench API

diasdavid:
Following the insights from https://github.com/ipfs/interface-ipfs-core/pull/99#issuecomment-264459732


diasdavid:
This lead to take in https://github.com/ipld/js-ipld-graph-builder as an API to work over graphs. It won't be part of the standard IPFS core, rather something that uses IPFS

PR #115, opened 2017-01-31
feat: update block spec to support CID

diasdavid:


diasdavid:
apparently, the spec was out of date in comparisson to the tests

Issue #122, opened 2017-03-09
dag api: get examples


jbenet:
### Example: `ipfs.dag.get`

```js
var IPFSAPI = require('ipfs-api')
var ipfs = IPFSApi('/ip4/127.0.0.1/tcp/5001')

// given an object such as: 
//
// var obj = {
//     "a": 1,
//     "b": [1, 2, 3],
//     "c": {
//         "ca": [5, 6, 7],
//         "cb": "foo"
//     }
// }
// 
// ipfs.dag.put(obj, function(err, cid2) {
//     assert(cid == zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y)
// })

function errOrLog(err, result) {
    if (err) console.error('error: ' + err)
    else console.log(result)
}

ipfs.dag.get('zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y/a', errOrLog)
// Returns:
// 1

ipfs.dag.get('zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y/b', errOrLog)
// Returns:
// [1, 2, 3]

ipfs.dag.get('zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y/c', errOrLog)
// Returns:
// {
//   "ca": [5, 6, 7],
//   "cb": "foo"
// }

ipfs.dag.get('zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y/c/ca/1', errOrLog)
// Returns:
// 6
```

diasdavid:
The difference here from what we have is that we expect the arguments to be `dag.get(cid, path, callback)`, where cid is an actual CID object and path is the string you want to search through. 

I can add support for this way too.

diasdavid:
The other difference is that the `result` has two properties:

```
{
  value: <the last resolved value>
  remainderPath: <the path missing to be resolved, if any, if not, empty string>
}
```

diasdavid:
Added these examples (modified a little to respect the actual API) on https://github.com/ipfs/interface-ipfs-core/pull/123

Issue #121, opened 2017-03-09
dag api: add {ls, tree}


jbenet:
We should add the following commands to the core/dag interface:

- `ipfs.dag.ls(path)` - show the first level link names of the node pointed at by `<path>`
  - include simple return value one, for small dag objects
  - include stream/iterator based one, for _huge_ dag objects
- `ipfs.dag.tree(path)` - enumerate all entries in the graph, parting from node at `<path>`. 
  - include simple return value one, for small subgraphs
  - stream/iterator based so that we can explore massive graphs with it.

cc @diasdavid @whyrusleeping

jbenet:
the ipld explorers will get worked on next week. would be great to have this in before that work starts. https://hackmd.io/EYVgHApgTDAsC0BDRIDG9bCgRia2E8EiqAbAAymqojkDMiQA

jbenet:
### Example for ipfs.dag.ls

```js
var IPFSAPI = require('ipfs-api')
var ipfs = IPFSApi('/ip4/127.0.0.1/tcp/5001')

// given an object such as: 
//
// var obj = {
//     "a": 1,
//     "b": [1, 2, 3],
//     "c": {
//         "ca": [5, 6, 7],
//         "cb": "foo"
//     }
// }
// 
// ipfs.dag.put(obj, function(err, cid2) {
//     assert(cid == zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y)
// })

ipfs.dag.ls('zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y', function(err, result) {
    for (var i in result.entries) {
        console.log(result.entries[i])
    }
})
// Returns:
// a
// b
// c

ipfs.dag.ls('zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y/c', function(err, result) {
    for (var i in result.entries) {
        console.log(result.entries[i])
    }
})
// Returns:
// ca
// cb

ipfs.dag.ls('zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y/c/ca', function(err, result) {
    for (var i in result.entries) {
        console.log(result.entries[i])
    }
})
// Returns:
// 0
// 1
// 2
```

jbenet:
### Example: `ipfs.dag.tree`

```js
[...]

PR #123, opened 2017-03-09
More DAG API goodness

diasdavid:
Given:
- https://github.com/ipfs/interface-ipfs-core/issues/122
- https://github.com/ipfs/interface-ipfs-core/issues/121

Tasks:
- [x] update dag.put to return CID
- [x] update dag.get to accept CID and CID+Path as String
- [x] dag.tree

Other:
- https://github.com/ipfs/interface-ipfs-core/issues/125


API/dag/README.md
```diff
@@ -38,3 +49,71 @@ If no `callback` is passed, a [promise][] is returned.
...
+
+##### `JavaScript` - ipfs.dag.ls(cid [, options], callback)
+
+- `cid` - can be one of the following:
+  - a [CID](https://github.com/ipfs/js-cid) instance.
+  - a CID in its String format (i.e: zdpuAkxd9KzGwJFGhymCZRkPCXtBmBW7mB2tTuEH11HLbES9Y)
+  - a CID in its String format concatenated with the path to be resolved
+- `options` - a object that might contain the following values:
+  - `stream` - bool - if set to true, it will return a Node.js Readable Stream.
+  - `pull` - bool - if set to true, it will return a pull-stream.
```
diasdavid: @dignifiedquire what is your take on this (options for streaming)? See https://github.com/ipfs/interface-ipfs-core/issues/121

dignifiedquire: I am not a big fan of this, if we want a streaming interface make it multiple methods, i.e.

- `.ls` callback, no streaming
- `.lsStream` node stream
- `.lsPull` pull stream

or sth like that

VictorBjelkholm: Isn't returning just data VS returning a nodejs stream VS returning a pull-stream just an implementation details of the javascript implemention of IPFS? How would this be applied to other languages where neither of these are available?

VictorBjelkholm: Removed comment about JS specific things, section is under JavaScript so makes sense.

diasdavid: @dignifiedquire understood. Right now it is PITA because we have places where 'xStream' is returning a pull-stream, other places where there is an option (really legacy) for `buffer: true` and a lot of internals that are pull-streams and the 'external' (ie js-ipfs) is Node.js Streams for the sake of not breaking API.

I would be down to just sit, define a language for this different options (or through options object or through different methods) and then apply as a golden rule throughout all the modules.

dignifiedquire: yes, we should try unify this api, agreed

diasdavid: Ok, I'll come up with a proposal then based on things we've discussed.

diasdavid: here you go https://github.com/ipfs/interface-ipfs-core/issues/126

src/dag.js
```diff
@@ -51,7 +51,7 @@ module.exports = (common) => {
         }
       })
 
-      describe('.put', () => {
+      describe.only('.put', () => {
```
dignifiedquire: please remove

API/dag/README.md
```diff
@@ -38,3 +55,83 @@ If no `callback` is passed, a [promise][] is returned.
...
+**Example:**
+
+```JavaScript
+// example obj
+const obj = {
+  a: 1,
+  b: [1, 2, 3],
+  c: {
+    ca: [5, 6, 7],
+    cb: 'foo"'
```
dignifiedquire: why the `"`?

API/dag/README.md
```diff
@@ -38,3 +55,83 @@ If no `callback` is passed, a [promise][] is returned.
...
+function errOrLog(err, result) {
+  if (err) {
+    console.error('error: ' + err)
+  } else {
+    console.log(result.value)
+  }
+}
+
+ipfs.dag.get('zdpuAmtur968yprkhG9N5Zxn6MFVoqAWBbhUAkNLJs2UtkTq5/a', errOrLog)
+// Returns:
```
dignifiedquire: logs, not returns

API/dag/README.md
diasdavid: typo :)

src/dag.js
diasdavid: 👍

API/dag/README.md
diasdavid: 👍

PR #135, opened 2017-04-04
Feat/dht

dignifiedquire:


diasdavid:
@dignifiedquire seems you had a very old copy of interface ipfs core

PR #139, opened 2017-05-10
testing(): add object.get tests for multihash string

kenshyx:
These tests will pass after https://github.com/ipfs/js-ipfs/pull/849 is merged

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -789,14 +789,14 @@ module.exports = (common) => {
           })
       })
 
-      it('object.get', () => {
+      it('object.get', (done) => {
         const testObj = {
           Data: new Buffer('get test object'),
           Links: []
         }
 
-        return ipfs.object.put(testObj).then((node1) => {
-          return ipfs.object.get(node1.multihash).then((node2) => {
+        ipfs.object.put(testObj).then((node1) => {
+          ipfs.object.get(node1.multihash).then((node2) => {
             // because js-ipfs-api can't infer if the
             // returned Data is Buffer or String
             if (typeof node2.data === 'string') {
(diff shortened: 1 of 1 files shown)


src/object.js
```diff
@@ -781,14 +789,14 @@ module.exports = (common) => {
...
+      it('object.get', () => {
         const testObj = {
           Data: new Buffer('get test object'),
           Links: []
         }
 
-        ipfs.object.put(testObj).then((node1) => {
-          ipfs.object.get(node1.multihash).then((node2) => {
+        return ipfs.object.put(testObj).then((node1) => {
+          return ipfs.object.get(node1.multihash).then((node2) => {
```
diasdavid: @kenshyx all of these tests are written to test the callback API. Check at the bottom the section to test de promises API.

We always test in depth through the Callback API.

src/object.js
```diff
@@ -797,7 +805,10 @@ module.exports = (common) => {
 
             expect(node1.data).to.deep.equal(node2.data)
             expect(node1.links).to.deep.equal(node2.links)
-            done()
+            // get object from ipfs multihash string
+            return ipfs.object.get(node1.toJSON().multihash).then((node2) => {
+              expect(node2).to.exist()
+            });
```
diasdavid: Make this a separate test.

src/object.js
kenshyx: I don't understand, those lines are under `promise API` :)

diasdavid:
Oh, I just realized I didn't read correctly where the changes were being added and in fact, you were doing the right thing! It seems that all of the other Promise API tests are finishing through calling done instead of returning a promise, which is bad because errors are not caught.

I'm sorry for the confusion @kenshyx, would you like to fix all the Promise API tests? If not I can do it as my redemption :)

kenshyx:
Yes, I'll do it.

kenshyx:
Let me know if there is something else I have to update @diasdavid

PR #140, opened 2017-05-24
Add links to the Pin API

enricomarino:


enricomarino:
since the pin API are spec'ed, we could add the links to the spec'ed API...

diasdavid:
keep all of this reviews coming @enricomarino, they are super appreciated :)

Issue #138, opened 2017-05-08
ipfs.files.cat path


jbenet:
Why does this not work?

```
ipfs.files.cat('/ipfs/<hash>/path')
```

- The js-ipfs api says only a multihash or a string-encoded multihash. 
- calling `cat` on a path is standard IPFS api. not doing this is confusing to users of ipfs.
- nothing here: https://github.com/ipfs/js-ipfs#api takes a path. this isn't right.
- looked for alternative functions or examples that did this for 15 minutes. could not figure it out. 👎

diasdavid:
This is now fully working, documented and tested in both js-ipfs and js-ipfs-api  🌟

PR #143, opened 2017-05-29
feat: add support for paths on get and cat

diasdavid:


diasdavid:
for https://github.com/ipfs/interface-ipfs-core/issues/138

PR #150, opened 2017-08-26
Swarm Tests - WIP

VictorBjelkholm:
Not mergable yet

### DIFF
--- a/src/swarm.js
+++ b/src/swarm.js
@@ -14,6 +14,7 @@ module.exports = (common) => {
   describe('.swarm', () => {
     let ipfsA
     let ipfsB
+    let factoryInstance
 
     before(function (done) {
       // CI takes longer to instantiate the daemon,
@@ -23,6 +24,7 @@ module.exports = (common) => {
 
       common.setup((err, factory) => {
         expect(err).to.not.exist()
+        factoryInstance = factory
         series([
           (cb) => {
             factory.spawnNode((err, node) => {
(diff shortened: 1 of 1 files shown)


VictorBjelkholm:
Depends on https://github.com/ipfs/js-ipfs/pull/960

VictorBjelkholm:
Two things I hit during writing these tests:

* For some reason, we cannot listen to port 0 twice somehow... Having `/ip4/127.0.0.1/tcp/0` is fine if it's just one time, but if we do the same, I would expect us to get assigned two random ports from the OS, but it only happens the first time, then the second listener is stuck on port 0
* Calling `ipfs.swarm.peers` in the callback of `ipfs.swarm.connect` does not show the peer you just connected to. It seems there is some async task that happens in the background of `ipfs.swarm.connect` after the callback has been called, instead of waiting for the connection to be fully established.

diasdavid:
> For some reason, we cannot listen to port 0 twice somehow... Having /ip4/127.0.0.1/tcp/0 is fine if it's just one time, but if we do the same, I would expect us to get assigned two random ports from the OS, but it only happens the first time, then the second listener is stuck on port 0

That is strange. Can you replicate the same behaviour using just TCP sockets in a simple Node.js script? 

> Calling ipfs.swarm.peers in the callback of ipfs.swarm.connect does not show the peer you just connected to. It seems there is some async task that happens in the background of ipfs.swarm.connect after the callback has been called, instead of waiting for the connection to be fully established.

This should not be the case, but it certainly happens in the 'dialing dance' inside libp2p-swarm. This work -- https://github.com/libp2p/js-libp2p-swarm/issues/24 -- would solve this. Can you make sure to add a test for this case (a failing one) commented out and pointing to that issue so that we know what's going on?

src/swarm.js
```diff
@@ -107,6 +109,140 @@ module.exports = (common) => {
...
+              (cb) => {
+                nodeA.swarm.connect(nodeBAddress, (err) => {
+                  expect(err).to.not.exist()
+                  cb()
+                })
+              },
+              (cb) => {
+                // Waiting to make sure nodes are connected
+                setTimeout(cb, 1000)
+              },
```
diasdavid: Is this necessary for nodeA.swarm.peers or for nodeB.swarm.peers?

src/swarm.js
```diff
@@ -107,6 +109,140 @@ module.exports = (common) => {
...
+          })
+          it('Connecting two peers with two addresses each', (done) => {
+            let nodeA
+            let nodeB
+            let nodeBAddress
+            const configA = getConfig([
+              // Need to explicitly declare the ports as 0 doesn't work with
+              // multiple swarm addresses apparently
+              // Not sure how to deal with the case if the ports are busy, this
+              // is introducing flakyness
```
diasdavid: This is really weird to me. See https://github.com/ipfs/interface-ipfs-core/pull/150#issuecomment-325125461

VictorBjelkholm:
> That is strange. Can you replicate the same behaviour using just TCP sockets in a simple Node.js script?

Made some small test program but no problem there. Made a test-case for libp2p-swarm though that is currently failing, you can see it here: https://github.com/libp2p/js-libp2p-swarm/pull/227

src/swarm.js
VictorBjelkholm: Test case to reproduce the issue: https://github.com/libp2p/js-libp2p-swarm/pull/227

diasdavid: Add a TODO here and create an issue to track this so that we can merge this PR as it is now.

diasdavid:
@VictorBjelkholm can you confirm that:

- [ ] This was tested against js-ipfs-api
- [ ] https://github.com/ipfs/js-ipfs/pull/960#issuecomment-326524425 is all that is needed for js-ipfs?

VictorBjelkholm:
@diasdavid yes to both of them

PR #153, opened 2017-09-01
Listen on API and Gateway

VictorBjelkholm:
The reason for this is that js-ipfs expects these to be multiaddrs, when
my expectation was that these could be null and therefore disable the
functionality


diasdavid:
Sounds good. I think the issue was that it was not fully tested and yet claimed to be. Lapses happen,   it is ok. As a golden rule to avoid them, it is important to be very rigorous about testing before getting others involved in the review :)

PR #155, opened 2017-10-06
feat: add progress bar tests

dryajov:


src/files.js
```diff
@@ -160,6 +182,52 @@ module.exports = (common) => {
           })
         })
 
+        it('add a nested dir as array with progress', (done) => {
+          // Needs https://github.com/ipfs/js-ipfs-api/issues/339 to be fixed
+          // for js-ipfs-api + go-ipfs
+          if (!isNode) { return done() }
```
VictorBjelkholm: Hm, why only non-node environment?

dryajov: hmm, good question - this is basically a copy of the above tests which carries this flag.

diasdavid: @VictorBjelkholm check the issue linked in the comments. go-ipfs started error'ing in v0.4.2

@dryajov mind trying running the dir tests (by removing the if clause) against latest go-ipfs and see if this is still an issue?

src/files.js
```diff
@@ -123,6 +123,28 @@ module.exports = (common) => {
...
+          }
+
+          ipfs.files.add(bigFile, {progress: handler}, (err, res) => {
+            expect(err).to.not.exist()
+            expect(res).to.have.length(1)
+            const file = res[0]
+            expect(file.hash).to.equal(expectedMultihash)
+            expect(file.path).to.equal(file.hash)
+            expect(progCount).to.equal(58)
+            expect(progress).to.equal(57792)
```
diasdavid: Is this the actual file size?

dryajov: I ran all the tests and all have passed, I think this is the actual size. I'll double check tho.

diasdavid: It should test for it. `.to.equal(sizeOfTheFile)`

diasdavid: 🔴 @dryajov this is now the last thing to do in order to get the whole thing merged, mind checking for the right size?

src/files.js
diasdavid: 🔴 @dryajov did you get to try this one too?

src/files.js
```diff
@@ -160,6 +182,52 @@ module.exports = (common) => {
...
+          }
+
+          ipfs.files.add(dirs, {progress: handler}, (err, res) => {
+            expect(err).to.not.exist()
+            const root = res[res.length - 1]
+
+            expect(root.path).to.equal('test-folder')
+            expect(root.hash).to.equal(expectedRootMultihash)
+            expect(progCount).to.equal(8)
+            expect(progress).to.equal(5)
```
diasdavid: This has to be wrong. How can a nested dir full of files have a `progress` of 5 bytes only?

diasdavid:
This also needs for the docs to be updated, ref: https://github.com/ipfs/js-ipfs-api/pull/604#issuecomment-335222266

src/files.js
dryajov: I did, it seems to still be an issue with the latest master.

src/files.js
dryajov: I had to switch attention to - https://github.com/ipfs/js-ipfs/pull/1036, since that would be a major missing part of the whole effort. I'll resume this as soon as I get that one squared out.

src/files.js
dryajov: That is indeed wrong, a fix is coming.

dryajov:
This is currently being _held_ by - https://github.com/ipfs/js-ipfs/pull/1036. Through we can probably merge this on its own and it would cover js-ipfs-api interaction with the go-ipfs progress functionality is missing from the js-ipfs http api which the above issue is addressing.

dryajov:
docs have been updated.

PR #158, opened 2017-10-18
fix: make tests consistent across js-ipfs/go-ipfs

dryajov:


src/files.js
```diff
@@ -130,10 +130,10 @@ module.exports = (common) => {
           let progress = 0
           const handler = (p) => {
             progCount += 1
-            progress += p
+            progress = p
```
diasdavid: Rename to accumProgress so that it is clear what it is measuring

PR #163, opened 2017-10-23
fixes example should get the whole tree

thisconnect:
JavaScript implementation would only log the tree for 'a' which would log []

Pretty sure this is an example bug

### DIFF
--- a/SPEC/DAG.md
+++ b/SPEC/DAG.md
@@ -148,7 +148,7 @@ function errOrLog(err, result) {
   if (err) {
     console.error('error: ' + err)
   } else {
-    console.log(result.value)
+    console.log(result)
   }
 }
 
--- a/SPEC/DAG.md
+++ b/SPEC/DAG.md
@@ -152,7 +152,7 @@ function errOrLog(err, result) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


thisconnect:
also the result is an array, should I update this?

```js
ipfs.dag.tree('zdpuAmtur968yprkhG9N5Zxn6MFVoqAWBbhUAkNLJs2UtkTq5', errOrLog)
// Logs:
// a
// b
// b/0
// b/1
// b/2
// c
// c/ca
// c/ca/0
// c/ca/1
// c/ca/2
// c/cb
```

In js-ipfs it is:

```js
// Logs:
// [ 'a',
//   'b',
//   'b/0',
//   'b/1',
//   'b/2',
//   'c',
//   'c/ca',
//   'c/ca/0',
//   'c/ca/1',
//   'c/ca/2',
//   'c/cb' ]
```

thisconnect:
dag.tree should return an array https://github.com/ipfs/interface-ipfs-core/blob/master/src/dag.js#L369-L372

diasdavid:
thanks for catching this @thisconnect :)

thisconnect:
My first time playing with js-ipfs, instead of js-ipfs-api + go-ipfs. It's great to have dag api :) works nicely so far.

diasdavid:
That's great feedback, thank you :D Let me know as you find other things :)

Issue #165, opened 2017-11-09
js-ipfs tests are failing


richardschneider:
```
module.js:515
    throw err;
    ^

Error: Cannot find module 'aegir/fixtures'
    at Function.Module._resolveFilename (module.js:513:15)
    at Function.Module._load (module.js:463:25)
    at Module.require (module.js:556:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (c:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\interface-ipfs-core\src\files.js:12:21)
    at Module._compile (module.js:612:30)
    at Object.Module._extensions..js (module.js:623:10)
    at Module.load (module.js:531:32)
    at tryModuleLoad (module.js:494:12)
    at Function.Module._load (module.js:486:3)
    at Module.require (module.js:556:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (c:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\interface-ipfs-core\src\index.js:4:17)
    at Module._compile (module.js:612:30)
    at Object.Module._extensions..js (module.js:623:10)
    at Module.load (module.js:531:32)
    at tryModuleLoad (module.js:494:12)
    at Function.Module._load (module.js:486:3)
    at Module.require (module.js:556:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (c:\Users\Owner\Documents\GitHub\js-ipfs\test\core\interface\block.js:4:14)
    at Module._compile (module.js:612:30)
    at Object.Module._extensions..js (module.js:623:10)
    at Module.load (module.js:531:32)
    at tryModuleLoad (module.js:494:12)
    at Function.Module._load (module.js:486:3)
    at Module.require (module.js:556:17)
    at require (internal/module.js:11:18)
    at Suite.describe (c:\Users\Owner\Documents\GitHub\js-ipfs\test\core\interface\interface.spec.js:8:3)
    at Object.create (C:\Users\Owner\AppData\Roaming\npm\node_modules\aegir\node_modules\mocha\lib\interfaces\common.js:112:19)
    at context.describe.context.context (C:\Users\Owner\AppData\Roaming\npm\node_modules\aegir\node_modules\mocha\lib\interfaces\bdd.js:44:27)
    at Object.<anonymous> (c:\Users\Owner\Documents\GitHub\js-ipfs\test\core\interface\interface.spec.js:7:1)
[...]

PR #166, opened 2017-11-09
 aegir is a dependency

richardschneider:
Issue #165

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   },
   "homepage": "https://github.com/ipfs/interface-ipfs-core#readme",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -30,6 +30,7 @@
   },
   "homepage": "https://github.com/ipfs/interface-ipfs-core#readme",
   "dependencies": {
+    "aegir": "^12.1.0",
```
diasdavid: Why change it from devDep to dep?

richardschneider: Please see issue #165.  Basically `src\files.js:12:21` depends on it.

PR #169, opened 2017-11-14
Pubsub message fields

richardschneider:
Rename to topicCIDs should be topicIDs.  See the [background discussion](https://github.com/ipfs/js-ipfs/issues/1068).

### DIFF
--- a/src/pubsub.js
+++ b/src/pubsub.js
@@ -122,7 +122,7 @@ module.exports = (common) => {
               expect(msg.data.toString()).to.equal('hi')
               expect(msg).to.have.property('seqno')
               expect(Buffer.isBuffer(msg.seqno)).to.be.eql(true)
-              expect(msg).to.have.property('topicCIDs').eql([topic])
+              expect(msg).to.have.property('topicIDs').eql([topic])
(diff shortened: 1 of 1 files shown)


richardschneider:
@diasdavid Needs PR https://github.com/libp2p/js-libp2p-floodsub/pull/49 released.  Then update dependencies.

src/pubsub.js
```diff
@@ -122,7 +122,7 @@ module.exports = (common) => {
               expect(msg.data.toString()).to.equal('hi')
               expect(msg).to.have.property('seqno')
               expect(Buffer.isBuffer(msg.seqno)).to.be.eql(true)
-              expect(msg).to.have.property('topicCIDs').eql([topic])
+              expect(msg).to.have.property('topicIDs').eql([topic])
```
diasdavid: This means that js-ipfs-api is also incorrect. Mind checking that?

richardschneider: Good call, already fixed in https://github.com/ipfs/js-ipfs-api/pull/627

richardschneider:
@diasdavid have you updated dependency to new the `floodsub`?

Issue #58, opened 2016-08-15
Cancelable requests AKA Context


diasdavid:
We need a way to be able to cancel calls that have to hit the network, either through a timeout or by a voluntary cancelation. 

JavaScript doesn't provide a good interface for this natively, we need to research or build one our own.

For context, read https://blog.golang.org/context which is what we use in Go

dignifiedquire:
Discussion around cancellation in the promise community:

- https://github.com/whatwg/fetch/issues/27
- https://github.com/tc39/proposal-cancelable-promises
- Cancellation in bluebird: http://bluebirdjs.com/docs/api/cancellation.html

jbenet:
Random experiment. it's not correct, just a draft. https://gist.github.com/jbenet/11671be17cbe91c67a00d9638d99a45d

dignifiedquire:
@diasdavid I can start doing some research and planning here, but I don't think it is realistic we can keep this in milestone 2

diasdavid:
Expanding on this a bit...

The main issue comes from the `Inversion of Control` nature that Callback APIs forces us to submit, where we can't be sure how our callback (Continuation) is going to be called (if called at all). In the other hand, it seems Promises seem to get this a little bit better by removing time out of the equation and enabling us to pass 'values' around, independently when they get resolved, and in addition, the native Promise implementation already has a mechanism for situations where multiple outcomes can happen, making sure that only one of the paths is resolved, the [Promise.race](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/race).

What this means, is that from the standpoint of the js-ipfs core caller, we can have:

```JavaScript
const doOperationP = new Promise((resolve, reject) => {
  // do whatever it is supposed to do and call resolve when it is done
})

let cancel 
const cancelP = new Promise((resolve, reject) => {
  cancel = () => {
    // Do the operations to rewind the operation
    reject()
  }
})
const timeoutP = new Promise((resolve, reject) => { 
    setTimeout(reject, constants.TIMEOUT)
})
Promise.race([
  doOpP,
  cancelP,
  timeoutP
])
```

This way, only one of the operations is called, keeping the control within outside to compose what we need that fits our expectations.

[...]

PR #170, opened 2017-11-17
fix: Add tests for .block promise API

haadcode:
This PR will add tests for the promise API of .block

I also added a test for callback block.rm() as it was missing

### DIFF
--- a/src/block.js
+++ b/src/block.js
@@ -26,7 +26,7 @@ module.exports = (common) => {
       // CI takes longer to instantiate the daemon,
       // so we need to increase the timeout for the
       // before step
-      this.timeout(20 * 1000)
+      this.timeout(60 * 1000)
 
       common.setup((err, factory) => {
         expect(err).to.not.exist()
@@ -140,7 +140,13 @@ module.exports = (common) => {
       })
 
       describe('.rm', function () {
-        it('by CID', (done) => {
+        // We need to skip this part of the test for now:
+        // By default, block.get() will try to get the block from
+        // bitswap thus causing a long timeout. What we want here is
+        // to get the block.get() to return an error as fast as possible,
+        // indicating that we don't have the block anymore.
+        // See: https://github.com/ipfs/interface-ipfs-core/pull/170#discussion_r151633508
+        it.skip('by CID', (done) => {
           const expectedHash = 'QmPv52ekjS75L4JmHpXVeuJ5uX2ecSfSZo88NSyxwA3rAQ'
           const cid = new CID(expectedHash)
           const b = new Block(new Buffer('blorb'), cid)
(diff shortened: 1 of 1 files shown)


haadcode:
Needed for https://github.com/ipfs/js-ipfs/pull/1085

src/block.js
```diff
@@ -137,8 +138,131 @@ module.exports = (common) => {
...
+              ipfs.block.rm(cid, (err) => {
+                expect(err).to.not.exist()
+                cb()
+              })
+            },
+            (cb) => {
+              // Remove bitswap from IPFS so that block.get tries to fetch the block
+              // from the local repo (not from the network). This is to make sure we
+              // get the error (as expected) instead of waiting for a timeout.
+              ipfs._blockService._bitswap = null
```
diasdavid: The tests can't use any internals otherwise it will fail with js-ipfs-api

haadcode: Oh :/ I figured that was a a bit of a workaround. Any suggestions how to avoid the block.get() trying to get the block from bitswap thus causing it "never" to return? The idea here is to get block.get() to return an error (ie. "we don't have the block anymore").

diasdavid: You will just have to .skip that test until we have https://github.com/ipfs/interface-ipfs-core/issues/58, unfortunately.

haadcode: Ok. That's unfortunate. I'll update the commit and .skip it for now. Let's leave the tests there though, so that we have them ready for the future.

haadcode: Committed. I removed the line where bitswap was nulled, added .skip to both callback and promise tests for .rm() and added some notes in the code as to why.

haadcode:
@diasdavid I believe these are running fine with both now after skipping the .rm() part. Did I miss something? Is there anything we need to change to get this merged?

diasdavid:
Hi @haadcode, this PR needs a rebase :)

diasdavid:
ping @haadcode

haadcode:
Sorry this took a while, it's rebased now! 🚀

src/block.js
```diff
@@ -38,108 +40,233 @@ module.exports = (common) => {
 
     after((done) => common.teardown(done))
 
-    describe('.put', () => {
-      it('a buffer, using defaults', (done) => {
-        const expectedHash = 'QmPv52ekjS75L4JmHpXVeuJ5uX2ecSfSZo88NSyxwA3rAQ'
-        const blob = new Buffer('blorb')
+    describe('callback API', () => {
```
diasdavid: @haadcode I believe you missed that there is no more separation between Callback API and Promise API and through your rebase you brought it back and duplicated some come.

haadcode: I did not notice that. Will fix it.

haadcode: @diasdavid I started to look into this but I'm little confused. Can you help here to elaborate what you mean with "no more separation between Callback and Promise API"? We still have them both, right?

diasdavid: We still have both, but we don't duplicate the tests anymore :)

diasdavid:
Closing this one.

PR #177, opened 2017-11-26
Prints the message content

joaosantos15:
In the subscribe and unsubscribe usage examples it now prints the message content rather than [Object object].


diasdavid:
Hi João, thanks. This PR assumes that the content of the message is always a string :)  Won't work for other cases.

PR #178, opened 2017-12-04
test: use os.tmpdir()

richardschneider:
See https://github.com/ipfs/js-ipfs/issues/1075

### DIFF
--- a/src/swarm.js
+++ b/src/swarm.js
@@ -11,6 +11,7 @@ const series = require('async/series')
 const multiaddr = require('multiaddr')
 const os = require('os')
 const path = require('path')
+const hat = require('hat')
 
 module.exports = (common) => {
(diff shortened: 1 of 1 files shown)


src/swarm.js
```diff
@@ -125,7 +127,7 @@ module.exports = (common) => {
           }
 
           function getRepoPath () {
-            return '/tmp/.ipfs-' + Math.random().toString().substring(2, 8) + Date.now()
+            return path.join(os.tmpdir(), '.ipfs-' + Math.random().toString().substring(2, 8) + Date.now())
```
diasdavid: While you are at it, could you do the same thing I requested at https://github.com/ipfs/ipfs-performance-profiling/pull/12#discussion_r154862400

PR #179, opened 2017-12-04
doc: modify javascripte usage link in README.md

qpakzk:
Because the link of javascript usage was wrong, it is correct to be changed to #javascript.


diasdavid:
Thanks @qpakzk :)

Issue #174, opened 2017-11-21
pubsub unsubscribe callback


richardschneider:
`unsubscribe` should allow promises and a callback, just like the other methods.

richardschneider:
A work around has been proposed https://github.com/ipfs/js-ipfs/issues/1137

PR #181, opened 2017-12-10
test(pubsub): subscribe with a Promise

richardschneider:
Here's a test for #174 that currently fails.


diasdavid:
Thank you @richardschneider ! :D

PR #182, opened 2017-12-12
specifies files.cat unknown file and dir errors

pgte:
Added 2 specs for `file.cat`:

- when the path is unknown
- when the path is a dir

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -431,6 +431,31 @@ module.exports = (common) => {
             }
           })
       })
+
+      it.only('errors on unknown path', () => {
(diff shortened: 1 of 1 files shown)


src/files.js
```diff
@@ -431,6 +431,31 @@ module.exports = (common) => {
             }
           })
       })
+
+      it.only('errors on unknown path', () => {
```
richardschneider: You should remove the `.only` and also on line 443

pgte: @richardschneider sorry, forgot to push.

diasdavid:
@pgte did you test this against js-ipfs and js-ipfs-api ?

PR #185, opened 2017-12-17
fix: key.rm test

richardschneider:
The key remove test was wrong.  This should fix https://github.com/ipfs/js-ipfs-api/pull/641

### DIFF
--- a/src/key.js
+++ b/src/key.js
@@ -119,7 +119,7 @@ module.exports = (common) => {
       })
 
       it('removes a key', function (done) {
-        ipfs.key.rm(key.name, (err, res) => {
+        ipfs.key.rm(key.Name, (err, res) => {
           expect(err).to.not.exist()
           expect(res).to.exist()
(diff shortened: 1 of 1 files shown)


src/key.js
```diff
@@ -133,7 +133,7 @@ module.exports = (common) => {
       it('does not contain the removed name', (done) => {
         ipfs.key.list((err, res) => {
           expect(err).to.not.exist()
-          const found = res.Keys.filter(k => k.Name === key.name)
+          const found = res.Keys.filter(k => k.Name === key.Name)
```
diasdavid: I don't disagree with the lowercase .name. You can convert that in js-ipfs-api. It is friendlier to js devs.

richardschneider: @diasdavid go-ipfs returns Capitalised names.  We must support it as is.

diasdavid: Fine then

PR #188, opened 2017-12-30
pubsub makes no guarantee on Windows and with go-ipfs

richardschneider:
Skip some tests because pubsub, `go floodsub`, does not make any guarantee on message delivery.

First reported in https://github.com/ipfs/js-ipfs-api/pull/648
```
1) .pubsub multiple nodes connected multiple nodes receive multiple messages:
     Error: Timeout of 80000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

2) .pubsub multiple nodes connected load tests send/receive 10k messages:
     Error: Timeout of 80000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

### DIFF
--- a/js/src/pubsub.js
+++ b/js/src/pubsub.js
@@ -13,6 +13,10 @@ const whilst = require('async/whilst')
 const each = require('async/each')
 const hat = require('hat')
 
+// On Browsers it will be false, but the tests currently aren't run
+// there anyway
+let isWindows = process.platform && process.platform === 'win32'
+
 function waitForPeers (ipfs, topic, peersToWait, callback) {
   const i = setInterval(() => {
     ipfs.pubsub.peers(topic, (err, peers) => {
@@ -67,6 +71,7 @@ module.exports = (common) => {
     let ipfs1
     let ipfs2
     let ipfs3
+    let withGo
 
     before(function (done) {
       // CI takes longer to instantiate the daemon, so we need to increase the
@@ -90,7 +95,12 @@ module.exports = (common) => {
           ipfs1 = nodes[0]
           ipfs2 = nodes[1]
           ipfs3 = nodes[2]
-          done()
+
+          ipfs1.id((err, id) => {
+            expect(err).to.not.exist()
+            withGo = id.agentVersion.startsWith('go-ipfs')
+            done()
+          })
         })
       })
     })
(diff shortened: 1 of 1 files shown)


diasdavid:
@richardschneider this looks like an incorrect assessment. The tests have shown only that **go-ipfs misses delivering messages to itself (having the same node be the publisher and the subscriber) only on Windows** which tells us that there might be some real issue and not just a protocol design. There have been more users that reported the same.

richardschneider:
Took a look at `go-floodsub` and [this code](https://github.com/libp2p/go-floodsub/blob/master/floodsub.go#L418) looks suspicious

> binary.BigEndian.PutUint64(seqno, uint64(time.Now().UnixNano())) 

Duplicate sequence number can be produced 
- if the resolution of the clock is not nanoseconds, or
- mutliple publish events are received within the resolution of the clock

diasdavid:
@richardschneider added a check to see if the test is being run with go-ipfs. However, we should also check if it is Windows given that the issue only happens there.

We can't use the `os` module as it wouldn't work in the Browser. Open to ideas :)

mkg20001:
@diasdavid Maybe aegir could set an environment variable like `AEGIR_OS` which could be included in the browser.

richardschneider:
@diasdavid The issue was not really Windows related but was related to the resolution of the system clock; see issue https://github.com/libp2p/go-floodsub/issues/52.

The fix is in the pipe-line.  So it should be resolved in the next go release.  

Let's merge this but  raise an issue to re-test when next go release is available.

diasdavid:
@Stebalien ETA for the next go-ipfs release?

richardschneider:
@diasdavid @Stebalien still waiting on the next go release, any estimate?

Stebalien:
"soon", unfortunately, this is a very big release. We need to get https://github.com/ipfs/go-ipfs/pull/4610 merged first, then test it, and then fix any fallout.

diasdavid:
@richardschneider let's do a skip on Windows + go-ipfs for these tests and add a `// TODO` to follow up after release

js/src/pubsub.js
```diff
@@ -13,6 +13,11 @@ const whilst = require('async/whilst')
 const each = require('async/each')
 const hat = require('hat')
 
+let isWindows = false
+if (process.platform && process.platform === 'win32') {
```
VictorBjelkholm:
```
const isWindows = process.platform && process.platform === 'win32'
```

Doesn't standard warn you here that this is a uneccessary conditional?

vmx: Good catch. I didn't get a warning but fixed it nonetheless.

PR #190, opened 2018-01-02
[WIP] docs: add examples to name API

hacdias:


hacdias:
@diasdavid, the spec for `ipfs.name.resolve` says you receive a response in this format:

```
{
  name: "/ipns/QmHash.."
  value: "/ipfs/QmHash.."
}
```

Although, what I am receiving here is something like:

```
{ Path: '/ipfs/QmQrX8hka2BtNHa8N8arAq16TCVx5qHcb46c5yPewRycLm' }
```

For `ipfs.name.publish`, I receive an object with capitalized fields. Instead of `name` and `value`, I receive a `Name` and a `Value`.

Since I'm adding some examples, should they be consistent with the spec or to the actual API? Should I make a PR on `js-ipfs-api` to change this?

hacdias:
@diasdavid one more thing: I see documentation all over the place. What's [ipfs/docs](https://github.com/ipfs/docs) supposed to be and how does it relate to [SPEC](https://github.com/ipfs/interface-ipfs-core/tree/master/SPEC) and to the docs on the other repositories? 😄

diasdavid:
ipfs/docs is a WIP not yet ready for prime time. Right now SPEC is the source of truth.

> Should I make a PR on js-ipfs-api to change this?

Thanks for catching this. Yes, please do make a PR :)

hacdias:
Done!

PR #192, opened 2018-01-03
Normalize KEY API

hacdias:
- [x] Normalize API with non-capitalized object fields
- [x] Wait for https://github.com/ipfs/js-ipfs-api/pull/659 to be completed.
- [x] Update Tests
 
Ref.: https://github.com/ipfs/js-ipfs-api/pull/659

### DIFF
--- a/SPEC/KEY.md
+++ b/SPEC/KEY.md
@@ -94,7 +94,8 @@ If no `callback` is passed, a promise is returned.
 ```JavaScript
 ipfs.key.rm('my-key', (err, key) => console.log(key))
 
-// { id: 'QmWETF5QvzGnP7jKq5sPDiRjSM2fzwzNsna4wSBEzRzK6W',
+// {
+//  id: 'QmWETF5QvzGnP7jKq5sPDiRjSM2fzwzNsna4wSBEzRzK6W',
 //   name: 'my-key' }
 ```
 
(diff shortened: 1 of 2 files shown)


hacdias:
@diasdavid just noticed that there is a `ipfs.key.export` in the SPEC but it is not a command on GO-IPS, nor in JS-IPFS, nor in the HTTP API.

Is it some TODO or should be removed?

hacdias:
@diasdavid done. Should be merged alongside with https://github.com/ipfs/interface-ipfs-core/pull/192.

hacdias:
@diasdavid done. Should be merged alongside with https://github.com/ipfs/interface-ipfs-core/pull/192.

richardschneider:
@hacdias PR https://github.com/ipfs/js-ipfs/pull/1133 has the implementation (core, cli and http) for the Key API.  Waiting for approval from @diasdavid 

With the case change, you will also need to change core to use lower case, because core and http-api must be compatible.  

However, http must still return Capitalized names because that is the spec defined by `go-ipfs`.  This allows a go client to use a js-ipfs daemon.

The cli can either talk directly to core or use http-api, so it must be changed to use lower case.

hacdias:
@richardschneider oops! Didn't notice that PR.

I just made this (and ipfs/js-ipfs-api#659) because most of the JS API is using non-capitalized object fields and that's just like the Standard specification tells us.

As far as I understood:

- HTTP API must return capitalized names (so `js-ipfs` and `go-ipfs` must return them that way).
- External APIs should be compliant to each language standard, which, in JS case, is to be non-capitalized.

Did I miss something?

richardschneider:
@hacdias I think we are on the same page.  When these `normalize`  PRs are released, please ping me and I'll update https://github.com/ipfs/js-ipfs/pull/1133

hacdias:
@diasdavid do you agree with what I [said](https://github.com/ipfs/interface-ipfs-core/pull/192#issuecomment-355126953)?

diasdavid:
thank you both for figuring this out and reaching consensus. LGTM :)

hacdias:
Thanks @diasdavid! Pinging @richardschneider 😄

Issue #193, opened 2018-01-04
Add files with progress callback - called 58 times?


alanshaw:
https://github.com/ipfs/interface-ipfs-core/blob/master/src/files.js#L115

Just wondering where there 58 comes from and what it's testing? I think it might be a brittle check as it's probably determined by data storage/transfer in much deeper dependencies that these tests have no control over.

It might work for now (and on the systems the tests are run on) but break later. I think it's enough to check that the progress function was called at least once and that the reported progress bytes is equal to the total bytes uploaded.

Would you accept a PR for this change?

VictorBjelkholm:
Indeed testing how many times the handler is called is testing the internal implementation rather than expecting correct output. I would happily accept a fix for this.

PR #194, opened 2018-01-04
Ensure progress handler was called, not the specific number of times

alanshaw:
As per the discussion in #193 this PR changes the tests involving the progress option to only test that the handler was called, not the number of times the handler was called.


RangerMauve:
Is there a test for making sure the progress cb doesn't get called after completion?

PR #196, opened 2018-01-08
Fixes timeout ms arg being erroneously passed to pubsub.peers 

alanshaw:
...and removes the timeout...it shouldn't be required

Also fixes a typo I noticed in the pubsub spec.

### DIFF
--- a/src/pubsub.js
+++ b/src/pubsub.js
@@ -298,16 +298,15 @@ module.exports = (common) => {
             (cb) => ipfs3.pubsub.subscribe(topicOther, sub3, cb)
           ], (err) => {
             expect(err).to.not.exist()
-            setTimeout(() => {
-              ipfs1.pubsub.peers(topic, (err, peers) => {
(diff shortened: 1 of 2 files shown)


src/pubsub.js
```diff
@@ -298,16 +298,15 @@ module.exports = (common) => {
...
-            setTimeout(() => {
-              ipfs1.pubsub.peers(topic, (err, peers) => {
-                expect(err).to.not.exist()
 
-                expect(peers).to.be.empty()
-                ipfs1.pubsub.unsubscribe(topic, sub1)
-                ipfs2.pubsub.unsubscribe(topicOther, sub2)
-                ipfs3.pubsub.unsubscribe(topicOther, sub3)
-                done()
-              }, 10000)
```
alanshaw: `10000` is being passed to `pubsub.peers` not to `setTimeout` (which I believe is what was intended)

PR #200, opened 2018-01-11
Fixes dns test with callback

alanshaw:
Test was previously passing because mocha didn't know it was async and no assertions were made


hacdias:
Thank you for fixing this! Really didn't notice 😅

PR #206, opened 2018-01-22
Spec MFS Actions

hacdias:
Spec MFS functions:

- ipfs.files.cp
- ipfs.files.mkdir
- ipfs.files.stat
- ipfs.files.rm
- ipfs.files.read
- ipfs.files.write
- ipfs.files.mv
- ipfs.files.flush(path, [callback])


hacdias:
@diasdavid just finished adding the specs for the MFS specific functions. Everything I wrote can be seen [from here](https://github.com/ipfs/interface-ipfs-core/blob/mfs/SPEC/FILES.md#cp).

SPEC/FILES.md
```diff
@@ -539,6 +539,252 @@ pull(
 
 A great source of [examples][] can be found in the tests for this API.
 
+#### `cp`
```
diasdavid: The MFS specific commands need a separator and an introduction to what MFS is.

hacdias: Okay @diasdavid. It makes sense. Although, there is a command (`ls`) who was already spec'ed and it says that it is only for MFS **but** that it should be implemented globally. [See](https://github.com/ipfs/interface-ipfs-core/blob/a62686ff8834dc3f7ac7bc4fde1c302bdf8dfc01/SPEC/FILES.md#javascript---ipfslsipfspath-callback).

hacdias:
@diasdavid just added a separator and a brief description.

diasdavid:
@hacdias updated wording. Seems you confused `ipfs.ls` and `ipfs.files.ls`. `ipfs.files.ls` is missing

hacdias:
Oh, it already said "files.ls" so I thought it would be the same. Thanks! I'll add the .Files.ls function today.

hacdias:
Done @diasdavid 😄

diasdavid:
Sweet!

@hacdias can you add tests for these two? You will have to enable them to go-ipfs only. You can do it by checking:

https://github.com/ipfs/interface-ipfs-core/blob/master/src/key.js#L33
https://github.com/ipfs/interface-ipfs-core/blob/master/src/key.js#L178

hacdias:
@diasdavid tests done! 😄 They seem to work on https://github.com/ipfs/js-ipfs-api/pull/675

src/files.js
```diff
@@ -980,5 +984,394 @@ module.exports = (common) => {
         )
       })
     })
+
+    describe('.mkdir', function () {
+      it('make directory on root', (done) => {
+        if (!withGo) {
+          console.log('Only supported by go-ipfs yet')
```
diasdavid: `Not supported in js-ipfs yet` would be better since those logs will only be printed when running js-ipfs tests.

src/files.js
```diff
@@ -980,5 +984,394 @@ module.exports = (common) => {
         )
       })
     })
+
```
diasdavid: Can we break the MFS tests into another file? This test file is huge :)

hacdias: Yup!

src/files.js
hacdias: Sure!

src/files.js
hacdias: Do you think it should be added to index.js as a separate tests suit or should we just import them in files.js and there we would export them both?

hacdias:
Done @diasdavid 😄 Check this: https://github.com/ipfs/interface-ipfs-core/pull/206/files#diff-ecdbeb0b77447177e8165fbaf62b5a09

PR #208, opened 2018-01-25
feat: STATS spec

hacdias:
Add STATS spec and tests! Ref.: https://github.com/ipfs/js-ipfs-api/pull/679


hacdias:
@diasdavid done 😄

PR #209, opened 2018-01-25
fix: stats not implemented on jsipfs

hacdias:
Ref.: https://github.com/ipfs/js-ipfs/pull/1189

### DIFF
--- a/src/stats.js
+++ b/src/stats.js
@@ -81,7 +81,7 @@ module.exports = (common) => {
     it('.bw', (done) => {
       if (!withGo) {
         console.log('Not supported in js-ipfs yet')
-        return
+        return done()
       }
 
       ipfs.stats.bw((err, res) => {
@@ -113,7 +113,7 @@ module.exports = (common) => {
     it('.repo', (done) => {
       if (!withGo) {
         console.log('Not supported in js-ipfs yet')
-        return
+        return done()
       }
 
       ipfs.stats.repo((err, res) => {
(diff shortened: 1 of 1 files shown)


hacdias:
@diasdavid could you check if [this](https://github.com/ipfs/interface-ipfs-core/pull/209/files#diff-6635c0bbe87eda053376dbd64f1dab6cR101) (only returning when we're using promises) is right?

hacdias:
@diasdavid just confirmed and it works.

src/stats.js
```diff
@@ -76,7 +96,12 @@ module.exports = (common) => {
     })
 
     it('.bw Promise', () => {
-      return ipfs.stats.bw().then((res) => {
+      if (!withGo) {
+        console.log('Not supported in js-ipfs yet')
+        return
+      }
+
+      ipfs.stats.bw().then((res) => {
```
diasdavid: @hacdias you should add a return here. Otherwise the test will pass before the call happens.

hacdias: OOPS! Sorry. fixed!

hacdias:
@diasdavid fixed 😄

src/stats.js
```diff
@@ -101,7 +101,7 @@ module.exports = (common) => {
         return
       }
 
-      ipfs.stats.bw().then((res) => {
+      return ipfs.stats.bw().then((res) => {
```
diasdavid: @hacdias have you checked if this is done in all the tests?

hacdias: Yes, on all promises tests

diasdavid: thank you 👍

PR #211, opened 2018-01-30
feat(breaking change): stats.bwReadableStream and stats.bwPullStream

### LINKED ISSUE #212: Many tests without actual assertions
A lot of tests are currently not actually doing any valuable assertions on the responses from calls. Rather, it just makes sure there is no errors and that keys exists.

Example: https://github.com/ipfs/interface-ipfs-core/blob/9d91267ab7ab61e41c84eaa81c647cd630d83797/js/src/stats.js#L39-L59

In this sample, `res` could have all those keys set to `null` and the test would pass. We should at least assert the type of the value, but best would be the exact value if we can. If not exact, we should be able to assert the format instead.

hacdias:
More details on https://github.com/ipfs/js-ipfs-api/pull/686

This also closes #212


hacdias:
Needed for https://github.com/ipfs/js-ipfs/pull/1198 /cc @diasdavid 😄

VictorBjelkholm:
@hacdias this should be a argument to the function that changes if it just returns the values once or if you get a stream back, rather than just having one of them.

The different signatures would be something like this:

```
ipfs.stats.bw((err, oneTimeRes) => {})
```

or

```
ipfs.stats.bw({poll: true}, () => {err, stream})
```

hacdias:
Sure! Makes sense. Will change that @VictorBjelkholm

hacdias:
@VictorBjelkholm done 😄

SPEC/STATS.md
```diff
@@ -25,7 +25,9 @@ Where:
   - `poll` is used to print bandwidth at an interval.
   - `interval` is the time interval to wait between updating output, if `poll` is true.
 
-`callback` must follow `function (err, stats) {}` signature, where `err` is an error if the operation was not successful. `stats` is an Object containing the following keys:
+If `poll` is `true`, then `callback` must follow `function (err, stream) {}` signature, where `err` is an error if the operation was not successful and `stream` is a Readable Stream where you can listen to the event `data` with a listener that must follow `function (stat) {}` signature.
+
+Otherwise, `callback` must follow `function (err, stat) {}` signature, where `err` is an error and `stat` is an Object containing the following keys:
```
VictorBjelkholm: We should define what the values are as well, I *think* the values are defined in bytes, but unsure.

hacdias: ? The `stat` thing on the stream?

hacdias: If so, it is an object like the other one. That's why I left them with the same name and explained in the end.

js/src/stats.js
```diff
@@ -95,6 +95,26 @@ module.exports = (common) => {
...
+      if (!withGo) {
+        console.log('Not supported in js-ipfs yet')
+        return done()
+      }
+
+      ipfs.stats.bw({poll: true}, (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+
+        res.once('data', (data) => {
```
VictorBjelkholm: We should have a way of closing the stream as well, for when you're done polling for stats

js/src/stats.js
```diff
@@ -95,6 +95,26 @@ module.exports = (common) => {
...
+          expect(data).to.have.a.property('totalIn')
```
VictorBjelkholm: This expectation only checks that the property exists, which I don't think is good enough. These test cases would still pass even if `totalIn` is set to for example `false`, `undefined` or even `null`, which is not correct.

VictorBjelkholm: Noticed that the two test cases are the same as well. One should test with `{poll: true}` and the other one without passing any extra args.

SPEC/STATS.md
VictorBjelkholm: Yeah, that `stat` object. We say which keys are on the object, but not what the values are.

VictorBjelkholm: Problem is not understanding that `stat` is used in both cases, but what the values are. Are they bytes? Are they kilobytes?

js/src/stats.js
hacdias: We have 4 test cases: 2 for callbacks and two for promises. One of those two has `{poll: true}` and the other has no args.

Right now, those values are strings. Should I convert them to something like [`big.js`](https://github.com/MikeMcl/big.js/)?

js/src/stats.js
```diff
@@ -110,6 +130,28 @@ module.exports = (common) => {
...
+          expect(data).to.have.a.property('totalOut')
+          expect(data).to.have.a.property('rateIn')
+          expect(data).to.have.a.property('rateOut')
+          done()
+        })
+      }).catch(err => {
+        expect(err).to.not.exist()
```
VictorBjelkholm: can just do `.catch(done)` instead as `done` with a error set will fail the test case

js/src/stats.js
hacdias: @VictorBjelkholm probably `stream.destroy()` is our best shot

js/src/stats.js
diasdavid: Since you are using Promises, you can remove the `done` and just make sure that you return the promise. The way it is now you don't get any benefit from using the Promise syntax over Callbacks

SPEC/STATS.md
```diff
@@ -25,7 +25,11 @@ Where:
   - `poll` is used to print bandwidth at an interval.
   - `interval` is the time interval to wait between updating output, if `poll` is true.
 
-`callback` must follow `function (err, stats) {}` signature, where `err` is an error if the operation was not successful. `stats` is an Object containing the following keys:
+If `poll` is `true`, then `callback` must follow `function (err, stream) {}` signature, where `err` is an error if the operation was not successful and `stream` is a Readable Stream where you can listen to the event `data` with a listener that must follow `function (stat) {}` signature.
```
diasdavid: Why ReadableStream and not pull-stream?

hacdias: I can use a pull-stream, but I didn't find a way [to do this](https://github.com/ipfs/js-ipfs-api/pull/686/files#diff-7b55c8c4d8fa3f0c172a6eaa8f2a80fdR39) with that package...

hacdias: Also, `js-ipfs-api` doesn't seem to be using pull streams. at least the package isn't a dependency.

SPEC/STATS.md
hacdias: I'll add that!

SPEC/STATS.md
hacdias: And how can we close a pull-stream? Their docs say nothing about it. The Readable Stream seems to be more straightforward and easy to use. You just have to do `.on('data', fn)`.

js/src/stats.js
hacdias: Added a `try catch` to check if they are numbers using `big.js.`

js/src/stats.js
hacdias: Added `res.destroy()` in the end.

hacdias:
@VictorBjelkholm I made some updates so this also fixes #212

hacdias:
@diasdavid, @VictorBjelkholm and I discussed about using Readable Stream or Pull Stream and he pointed me to https://github.com/ipfs/js-ipfs/issues/557#issuecomment-345372286. Then we noticed that it doesn't make much sense to have three different functions for this API since it is really simple.

Right now, I have made the implementation to use [Readable Stream here](https://github.com/ipfs/js-ipfs-api/pull/686/files#diff-7b55c8c4d8fa3f0c172a6eaa8f2a80fdR39). Do you think we should use `pull-stream`? If so, why and could you point me out some places to look to see how I should implement this because their docs didn't help me much.

diasdavid:
@hacdias there are some more bits that are important for you to get familiar with. Glad that you found https://github.com/ipfs/js-ipfs/issues/557, make sure to also:

- read through https://github.com/ipfs/js-ipfs/issues/362
- Watch https://www.youtube.com/watch?v=g5ewQEuXjsQ
- read http://dominictarr.com/post/145135293917/history-of-streams
- read http://dominictarr.com/post/149248845122/pull-streams-pull-streams-are-a-very-simple
- see how we support multiple streams interfaces at https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md
- understand the memory implications of https://gist.github.com/dominictarr/33042821928753986138942d0fdcefed
- get up to speed on https://github.com/ipfs/community/issues/278 and learn how this kind of polymorphism can have a severe perf impact.

> Then we noticed that it doesn't make much sense to have three different functions for this API since it is really simple.

I feel that the argument that you are trying to make is that "this is not a very used function and so a simple thing will though", which can be fair, however, it is important to understand the future implications and make decisions that will enable users to migrate without having to follow a changing interface. Two cery good threads for you to read: https://unix.stackexchange.com/questions/235335/the-linux-kernel-breaking-user-space & https://felipec.wordpress.com/2013/10/07/the-linux-way/

Let's follow up once you went through all of these materials :)

js/src/repo.js
```diff
@@ -46,26 +56,36 @@ module.exports = (common) => {
...
+        expect(res).to.have.a.property('numObjects')
+        expect(res).to.have.a.property('repoSize')
+        expect(res).to.have.a.property('repoPath')
+        expect(res).to.have.a.property('version')
+        expect(res).to.have.a.property('storageMax')
+        expect(isNumber(res.numObjects)).to.eql(true)
+        expect(isNumber(res.repoSize)).to.eql(true)
+        expect(isNumber(res.storageMax)).to.eql(true)
+        expect(res.repoPath).to.be.a('string')
+        expect(res.version).to.be.a('string')
```
VictorBjelkholm: This is still not valid as `expect('').to.be.a('string')` would pass the tests.

js/src/stats.js
```diff
@@ -54,6 +64,16 @@ module.exports = (common) => {
         expect(res).to.have.a.property('dataSent')
         expect(res).to.have.a.property('dupBlksReceived')
         expect(res).to.have.a.property('dupDataReceived')
+
+        expect(isNumber(res.provideBufLen)).to.eql(true)
+        expect(res.wantlist).to.be.an('array')
+        expect(res.peers).to.be.an('array')
```
VictorBjelkholm: Would we consider this test to be correct if `res.peers` was a empty array?

js/src/repo.js
```diff
@@ -6,8 +6,18 @@
 const chai = require('chai')
 const dirtyChai = require('dirty-chai')
 const expect = chai.expect
+const Big = require('big.js')
 chai.use(dirtyChai)
 
+const isNumber = (n) => {
```
VictorBjelkholm: This is actually checking if `n` is a BigInt, not `Number`, which is a built-in type in JavaScript

js/src/repo.js
```diff
@@ -46,26 +56,36 @@ module.exports = (common) => {
...
-      ipfs.repo.stat((err, stat) => {
+      ipfs.repo.stat((err, res) => {
         expect(err).to.not.exist()
-        expect(stat).to.exist()
-        expect(stat).to.have.property('numObjects')
-        expect(stat).to.have.property('repoSize')
-        expect(stat).to.have.property('repoPath')
-        expect(stat).to.have.property('version')
-        expect(stat).to.have.property('storageMax')
+        expect(res).to.exist()
```
VictorBjelkholm: These assertions are repeated all over the tests. We should extract the assertions into a function and pass in `res` instead.

```
const expectIsStat = (stat) => {
  // Make all the assertions once here
}
```

SPEC/STATS.md
```diff
@@ -25,12 +25,16 @@ Where:
...
 
-- `totalIn`
-- `totalOut`
-- `rateIn`
-- `rateOut`
+Otherwise, `callback` must follow `function (err, stat) {}` signature, where `err` is an error.
+
+`stat` is, in both cases, an Object containing the following keys:
+
+- `totalIn` - is a string containing a uint64 in bytes.
```
VictorBjelkholm: These are now [`Big`](https://github.com/MikeMcl/big.js/) values rather than strings.

js/src/stats.js
hacdias: Yes and it might be empty in some cases.

SPEC/STATS.md
```diff
@@ -25,12 +25,16 @@ Where:
   - `poll` is used to print bandwidth at an interval.
   - `interval` is the time interval to wait between updating output, if `poll` is true.
 
-`callback` must follow `function (err, stats) {}` signature, where `err` is an error if the operation was not successful. `stats` is an Object containing the following keys:
+If `poll` is `true`, then `callback` must follow `function (err, stream) {}` signature, where `err` is an error if the operation was not successful and `stream` is a Pull Stream where you can listen to the event `data` with a listener that must follow `function (stat) {}` signature.
```
diasdavid: @hacdias this wasn't necessarily what I was hoping for here. Follow the example of https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#files-api and create two api calls.

- `ipfs.stats.bw` - single requests
- `ipfs.stats.bwPoll` - return a stream

hacdias: `bwPoll` returns which _kind_ of stream?

diasdavid: - `ipfs.stats.bwPollPullStream`
- `ipfs.stats.bwPollReadableStream`

🙂

hacdias: I'd just suggest to take the 'Poll' out of the equation and use this three functions:

- `stats.bw`
- `stats.bwPullStream`
- `stats.bwReadableStream`

hacdias: What do you think @diasdavid?

hacdias: image

I suggest to structure the SPEC like that instead of creating different entries for each function and add examples for each of them. In FILES spec, we have:

image

Which is not accurate: we have one API endpoint which is `cat`. The other functions should be called variants or something like that and not separated in the SPEC. Why? Duplicated information (on the parameters for example) and then we have that `GO - WIP` thing which may make users think there will be `catPullStream` or `[insert your command here]ReadableStream` on Go. That's something I'd like to work on after finishing this Pull Request and make this interface documentation a little better. What do you think @diasdavid?

diasdavid: @hacdias let's avoid painting the 🚲🏡 in this PR. tl;dr; they are top level functions, they should be listed in the table of contents and have urls to it.

hacdias:
@diasdavid could you please take a look at the SPEC I wrote? If you approve it, I can start implementing it on `js-ipfs-api` and create the tests here 😄

diasdavid:
@hacdias go for it :)

hacdias:
I think this is ready @diasdavid :)

Issue #97, opened 2016-11-25
Standardize `Bootstrap API`


diasdavid:
- [x] spec
- [x] tests
- [x] run on js-ipfs
- [x] run on js-ipfs-api

diasdavid:
PRs:
- https://github.com/ipfs/js-ipfs/pull/1218
- https://github.com/ipfs/js-ipfs-api/pull/692

PR #220, opened 2018-02-12
Add bootstrap API spec

richardschneider:
Document the spec as implemented by `js-ipfs`. See #97


richardschneider:
@diasdavid When the spec is merged, I'll start writing the tests!

diasdavid:
@richardschneider you can use the tests from https://github.com/ipfs/js-ipfs-api/blob/master/test/bootstrap.spec.js just by importing them here.

It would be good to merge the spec with the tests and ensure that both js-ipfs and js-ipfs-api follow it so that we don't confuse users.

richardschneider:
Sounds like a good plan.  Will do tomorrow.

richardschneider:
@diasdavid As requested, I've added the tests.

PR #221, opened 2018-02-20
Adds missing docs for dht.provide

alanshaw:
As far as I can tell from the [tests](https://github.com/ipfs/interface-ipfs-core/blob/f5f73070c37f374e0516185b8241622d514d09ae/js/src/dht.js#L124) and [`js-ipfs`](https://github.com/ipfs/js-ipfs/blob/e4d2a15143431a9af37a275a149621738c91aa00/src/core/components/dht.js#L119) (follows through to [`js-ipfs-repo`](https://github.com/ipfs/js-ipfs-repo/blob/c30a7f170a631cc6298c06340bd92321b6fe1cd7/src/blockstore.js#L136) `.has` which expects a CID) it just takes a CID. `js-ipfs` implementation also suggests it can [take an array](https://github.com/ipfs/js-ipfs/blob/e4d2a15143431a9af37a275a149621738c91aa00/src/core/components/dht.js#L109).

### DIFF
--- a/SPEC/DHT.md
+++ b/SPEC/DHT.md
@@ -71,6 +71,28 @@ ipfs.dht.get(key, function (err, value) {})
 
 A great source of [examples][] can be found in the tests for this API.
 
+#### `provide`
+
+> Announce to the network that you are providing given values.
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.dht.provide(cid, [callback])
+
+Where `cid` is a CID or array of CIDs.
+
+`callback` must follow `function (err) {}` signature, where `err` is an error if the operation was not successful.
+
+If no `callback` is passed, a promise is returned.
+
+**Example:**
+
+```JavaScript
+ipfs.dht.provide(cid, function (err) {})
+```
+
+A great source of [examples][] can be found in the tests for this API.
+
 #### `put`
 
 > Store a value on the DHT
--- a/js/src/dht.js
+++ b/js/src/dht.js
@@ -118,21 +118,37 @@ module.exports = (common) => {
 
     describe('.provide', () => {
       it('regular', (done) => {
-        // TODO recheck this test, should it provide or not if block is not available? go-ipfs does provide.
+        nodeC.files.add(Buffer.from('test'), (err, res) => {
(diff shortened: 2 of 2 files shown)


diasdavid:
@alanshaw mind adding tests for this?

diasdavid:
ping @alanshaw :)

alanshaw:
haha, I'm doing this right now!

alanshaw:
There's not a lot to test here since it doesn't return any values, nevertheless I've added a test for passing multiple CIDs in an array asserting the call did not fail and I've added a test for passing a non CID and an array of non CIDs.

DHT tests are not enabled in `js-ipfs`. If I enable them (for node env) then they pass \o/.

The new DHT tests in `js-ipfs-api` fail:

screen shot 2018-03-06 at 16 50 07

alanshaw:
ah wow, so, this became more work than I bargained for 🙄 haha

**I realised that go-ipfs _doesn't_ provide if you ask it to provide a cid it doesn't have locally.**

The `js-ipfs-api` call to `dht/provide` was creating a querystring like:

screen shot 2018-03-06 at 20 43 41

I **think** go-ipfs isn't deserialize that `arg` properly and thinks you called `provide` with no `arg`, which is fine...it just succeeds. I believe this led to the incorrect assertion that "`go-ipfs` just provides".

I changed js-ipfs-api to serialize the cid to a base58 string _instead_ of letting `qs` serialize it (`qs` calls `toJSON`) and go-ipfs started erroring:

screen shot 2018-03-06 at 20 55 38

PR for that on its way.

I've also got a PR on its way for js-ipfs to reinstate the check for providing non-local cids which was commented out!

Anyway, with both those mentioned PRs the new tests in this PR now pass.

Have a nice day 🌷

alanshaw:
I've added a test for CIDv1 - it seems that go-ipfs is ok with this version of CID (no error), but the check for localness returns false in js-ipfs so it fails (although the actual call to `libp2pNode.contentRouting.provide` does not error if we remove the local check).

You wanna me to look into the localness issue?

diasdavid:
//cc @vmx

PR #223, opened 2018-02-22
fix: adapt dag tests to current environment

vmx:
The current go-ipfs doesn't yet support running all those tests,
but at least part of them. Fix the tests that work and skip the
ones that are not yet supported.

### DIFF
--- a/js/src/dag.js
+++ b/js/src/dag.js
@@ -255,7 +255,7 @@ module.exports = (common) => {
           }
           ipfs.dag.get(cidPb, 'Data', (err, result) => {
             expect(err).to.not.exist()
-            expect(result.value.data).to.eql(Buffer.from('I am inside a Protobuf'))
+            expect(result.value).to.eql(Buffer.from('I am inside a Protobuf'))
             done()
           })
         })
--- a/js/src/dag.js
+++ b/js/src/dag.js
@@ -17,6 +17,7 @@ const CID = require('cids')
 module.exports = (common) => {
   describe('.dag', () => {
     let ipfs
+    let withGo
 
     before(function (done) {
       // CI takes longer to instantiate the daemon, so we need to increase the
@@ -28,7 +29,11 @@ module.exports = (common) => {
         factory.spawnNode((err, node) => {
           expect(err).to.not.exist()
           ipfs = node
-          done()
+          ipfs.id((err, id) => {
+            expect(err).to.not.exist()
+            withGo = id.agentVersion.startsWith('go-ipfs')
+            done()
+          })
         })
       })
     })
@@ -67,17 +72,13 @@ module.exports = (common) => {
         }, done)
       })
 
-      /*
-       * This works because dag-cbor will just treat pbNode as a regular object
+      // This works because dag-cbor will just treat pbNode as a regular object
       it.skip('dag-pb node with wrong multicodec', (done) => {
-        // This works because dag-cbor will just treat pbNode as a
-        // regular object
         ipfs.dag.put(pbNode, 'dag-cbor', 'sha3-512', (err) => {
           expect(err).to.exist()
           done()
         })
       })
-      */
 
       it('dag-cbor with default hash func (sha2-256)', (done) => {
         ipfs.dag.put(cborNode, {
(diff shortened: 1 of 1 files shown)


vmx:
No need for review at the moment. I just figured out that they are also run by js-ipfs. I need to accommodate that fact.

vmx:
Ready for review.

js/src/dag.js
```diff
@@ -93,7 +94,10 @@ module.exports = (common) => {
         }, done)
       })
 
-      it('dag-cbor node with wrong multicodec', (done) => {
+      // This works because dag-pb will serialize any object. If the object
+      // has neither a `data` nor `links` field it's serialized as an empty
+      // object
+      it.skip('dag-cbor node with wrong multicodec', (done) => {
```
diasdavid: Why skip?

js/src/dag.js
```diff
@@ -106,7 +110,7 @@ module.exports = (common) => {
       it('returns the cid', (done) => {
         ipfs.dag.put(cborNode, {
           format: 'dag-cbor',
-          hashAlg: 'sha3-512'
+          hashAlg: 'sha2-256'
```
diasdavid: Why change? This should continue to work.

js/src/dag.js
```diff
@@ -239,10 +243,17 @@ module.exports = (common) => {
...
+          // TODO vmx 2018-02-22: Currently not supported in go-ipfs, it might
+          // be possible once https://github.com/ipfs/go-ipfs/issues/4728 is
+          // done
+          if (withGo) {
+            this.skip()
+          }
           ipfs.dag.get(cidPb, 'Data', (err, result) => {
             expect(err).to.not.exist()
-            expect(result.value).to.eql(Buffer.from('I am inside a Protobuf'))
+            console.log('vmx: result', result.value)
```
diasdavid: rm console.log

js/src/dag.js
vmx: Sorry, that's a left-over, it should only skip in case of go-ipfs (there the comment is accurate).

js/src/dag.js
vmx: The problem is that custom hash algorithms never worked: https://github.com/ipld/js-ipld/blob/6693f89d7970002ccdc209a5c2ad5b23edf23f84/src/index.js#L233-L236 So I decided to adapt the test rather than going into the rabbit hole of fixing it.

diasdavid: 👍

js/src/dag.js
```diff
@@ -239,10 +246,16 @@ module.exports = (common) => {
...
+          // TODO vmx 2018-02-22: Currently not supported in go-ipfs, it might
+          // be possible once https://github.com/ipfs/go-ipfs/issues/4728 is
+          // done
+          if (withGo) {
+            this.skip()
+          }
           ipfs.dag.get(cidPb, 'Data', (err, result) => {
             expect(err).to.not.exist()
-            expect(result.value).to.eql(Buffer.from('I am inside a Protobuf'))
+            expect(result.value.data).to.eql(Buffer.from('I am inside a Protobuf'))
```
diasdavid: Is this correct? Looks like `.data` should not be necessary.

vmx: Good catch. This time I actually ran the tests on js-ipfs.

vmx:
@diasdavid I think I've addressed everything, please re-review.

js/src/dag.js
```diff
@@ -312,7 +337,12 @@ module.exports = (common) => {
       })
     })
 
-    describe('.tree', () => {
+    describe.skip('.tree', function () {
```
diasdavid: Why skip for .all?

vmx: Sorry again :-/ Something is wrong with this PR. I ran it with both js-ipfs and js-ipfs-api. But after this change I will run it again.

vmx:
@diasdavid I did run the current code (with the most recent fix) on js-ipfs and on js-ipfs-api (one time on master and one time with my changes for the DAG API which will then actually run the test). All passed as expected.

PR #227, opened 2018-02-28
complete files.stat with the 'with-local' option

MichaelMure:
Following https://github.com/ipfs/go-ipfs/pull/4638, this new option is present in go-ipfs that enable to know if and how much of a dag is present locally.

Now, in the go-ipfs's http API, `WithLocality`, `Local` and `SizeLocal` in the answer are not present when `with-local` is not active. I chose to set default values in the javascript object response, but it could also be done the same way with no values present at all. Which would be best ?

### DIFF
--- a/SPEC/FILES.md
+++ b/SPEC/FILES.md
@@ -618,6 +618,7 @@ Where:
 - `options` is an optional Object that might contain the following keys:
   - `hash` is a Boolean value to return only the hash.
   - `size` is a Boolean value to return only the size.
+  - `withLocal` is a Boolean value to compute the amount of the dag that is local, and if possible the total size.
 
 `callback` must follow the `function (err, stat) {}` signature, where `err` is an Error if the operation was not successful and `stat` is an Object with the following keys:
 
@@ -626,6 +627,10 @@ Where:
 - `cumulativeSize` is an integer with the cumulative size in Bytes.
 - `blocks` is an integer indicating the number of blocks.
 - `type` is a string that can be either `directory` or `file`.
+- `withLocality` is a boolean to indicate if locality information are present.
+- `local` is a boolean to indicate if the queried dag is fully present locally.
+- `sizeLocal` is an integer indicating the cumulative size of the data present locally.
+
 
 If no `callback` is passed, a promise is returned.
 
--- a/js/src/files-mfs.js
+++ b/js/src/files-mfs.js
@@ -276,7 +276,10 @@ module.exports = (common) => {
             blocks: 1,
             size: 13,
             hash: 'QmcZojhwragQr5qhTeFAmELik623Z21e3jBTpJXoQ9si1T',
-            cumulativeSize: 71
+            cumulativeSize: 71,
+            withLocality: false,
+            local: undefined,
+            sizeLocal: undefined
           })
           done()
         })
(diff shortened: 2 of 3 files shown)


SPEC/FILES.md
```diff
@@ -618,6 +618,7 @@ Where:
 - `options` is an optional Object that might contain the following keys:
   - `hash` is a Boolean value to return only the hash.
   - `size` is a Boolean value to return only the size.
+  - `with-local` is a Boolean value to compute the amount of the dag that is local, and if possible the total size.
```
diasdavid: When options have `-`, we translate them to camel case. `withLocal`

MichaelMure: Will do, but it's currently not respected in js-ipfs-api. For instance: https://github.com/ipfs/js-ipfs-api/blob/master/test/files.spec.js#L93

diasdavid: Seems that one passed by review. That said, we still want to keep the camelCase version and support the dash for backwards compatibility

MichaelMure: I see that lodash is used exactly once in js-ipfs-api. Nothing against using `opts = _.mapKeys(opts, (v, k) => _.kebabCase(k))` ?

That would support both version and some more easily.

js/src/files-mfs.js
```diff
@@ -295,7 +298,54 @@ module.exports = (common) => {
...
+            sizeLocal: undefined
+          })
+          done()
+        })
+      })
+
+      it('stat withLocal file', (done) => {
+        if (!withGo) {
+          console.log('Not supported in js-ipfs yet')
+          return done()
```
diasdavid: call `this.skip()`. You will have to replace the arrow function by regular function

js/src/files-mfs.js
```diff
@@ -295,7 +298,54 @@ module.exports = (common) => {
...
+          })
+          done()
+        })
+      })
+
+      it('stat withLocal dir', (done) => {
+        if (!withGo) {
+          console.log('Not supported in js-ipfs yet')
+          return done()
+        }
```
diasdavid: same as above.

js/src/files.js
```diff
@@ -980,5 +985,31 @@ module.exports = (common) => {
...
       })
     })
+
+    describe('.stat', () => {
+      before((done) => ipfs.files.add(smallFile.data, done))
+
+      it('stat outside of mfs', () => {
+        if (!withGo) {
+          console.log('Not supported in js-ipfs yet')
+          return done()
```
diasdavid: same as above.

js/src/files-mfs.js
MichaelMure: This is the way it's done in all the codebase. Do you prefer to change that here or in a single pass in another PR ?

diasdavid: can you change the others first in a PR and update this one to match? thank you :)

MichaelMure:
@diasdavid I changed for `this.skip()` as well so it should be fine now

vmx:
This PR broke most of the `.stat` tests run by js-ipfs-api: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs-api/detail/get-ci-green/1/tests/

@MichaelMure Can you please have a look?

diasdavid:
@vmx you will need https://github.com/ipfs/js-ipfs-api/pull/695. Wanna update deps there? 

This is one of the reasons why I asked you for a issue that lists all the PR to make js-ipfs-api tests. There are so many PR's in flight that is hard to manage.

MichaelMure:
@vmx Thank for taking care of that.

vmx:
@MichaelMure  It wasn't me, @diasdavid fixed those tests.

MichaelMure:
Errr, thanks @diasdavid ^^

PR #229, opened 2018-03-06
test: Improve pubsub test isolation

vmx:
Always unsubscribe the feeds. This way subsequent tests won't
fail randomly.

### DIFF
--- a/js/src/pubsub.js
+++ b/js/src/pubsub.js
@@ -560,64 +560,76 @@ module.exports = (common) => {
           )
         })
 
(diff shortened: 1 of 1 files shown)


vmx:
I've run the full js-ipfs-api and js-ipfs test suite on Linux. js-ipfs-api passes, js-ipfs failed, but every re-run on different tests, so I don't think it's connected to this change.

As I want to keep things moving, @diasdavid please merge this one (I don't have permission).

PR #230, opened 2018-03-07
fix: docs and adds test assertion that peer is a PeerId in return value from swarm.peers

alanshaw:
As per discussion in https://github.com/ipfs/js-ipfs/issues/1248


diasdavid:
Is there a PR to fix this in js-ipfs?

alanshaw:
https://github.com/ipfs/js-ipfs/pull/1252

alanshaw:
Note, there's no change to js-ipfs-api as it already assumes a PeerId is being sent:
https://github.com/ipfs/js-ipfs-api/blob/master/src/swarm/peers.js#L48

PR #232, opened 2018-03-07
test: make fixtures cross-platform compatible

vmx:
Git on Windows changes the end of lines from LF to CRLF by default.
This changes the bytes of fixtures that a are text files. Change
the Git repository setting to checkout all files in `test/fixtures`
as they are, without platform specific modifications


VictorBjelkholm:
Rather than having this change basically everywhere we have fixtures, it might be better to just document that if you want to run tests on windows for js projects, you'll have to configure Git to use input crlf instead of default.

Basically, user's dev env should run `git config --global core.autocrlf input` before cloning the repositories.

What do you think?

JonKrone:
I actually ran into this and a similar issue where levelDB fails because of line-endings in fixtures during my first week of development: https://github.com/ipfs/js-ipfs/issues/1204#issuecomment-367064044. It caused me to switch the issue I was working on because I couldn't figure it out in a couple hours. There are so many points of entry to js-ipfs land that it is hard to ensure that developers will see a warning. While I also don't like adding this to all fixtures, this problem has a broad indirect impact and reduces friction for newcomers.

vmx:
@VictorBjelkholm: Having two different workflows for running tests and not seems counter-intuitive. I would rather add `.gitattributes` to all fixtures. I even have writing a script that checks all repos which have text files in there and make a pull request with a `.gitattributes` file on my TODO list.

PR #234, opened 2018-03-09
fix: files.add accepts object

dryajov:


diasdavid:
@dryajov have you run these against js-ipfs and js-ipfs-api? PRs for both?

PR #235, opened 2018-03-09
fix: don't error to specific

vmx:
When stopping an IPFS node it returns an error. The error message
depends on the environment. In Browsers it might also return
`Failed to fetch` or `XHR error`. Hence removing the specific
error check as it would fail for Browsers.

### DIFF
--- a/js/src/miscellaneous.js
+++ b/js/src/miscellaneous.js
@@ -11,6 +11,7 @@ chai.use(dirtyChai)
 module.exports = (common) => {
   describe('.miscellaneous', () => {
     let ipfs
+    let withGo
 
     before(function (done) {
       // CI takes longer to instantiate the daemon, so we need to increase the
@@ -22,7 +23,11 @@ module.exports = (common) => {
         factory.spawnNode((err, node) => {
           expect(err).to.not.exist()
           ipfs = node
-          done()
+          ipfs.id((err, id) => {
+            expect(err).to.not.exist()
+            withGo = id.agentVersion.startsWith('go-ipfs')
+            done()
+          })
         })
       })
     })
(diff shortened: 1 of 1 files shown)


vmx:
I'm not sure if that's the proper fix. I hope @richardschneider would know.

vmx:
It's not the proper fix. It also failed for me. I need to revisit this on Monday.

vmx:
Sorry for the noise, with increasing the timeout it seems to now pass all the time.

js/src/miscellaneous.js
```diff
@@ -83,12 +83,11 @@ module.exports = (common) => {
     })
 
     // must be last test to run
-    it('.stop', (done) => {
+    it('.stop', function (done) {
+      this.timeout(10 * 1000)
       // TODO: go-ipfs returns an error, https://github.com/ipfs/go-ipfs/issues/4078
       ipfs.stop((err) => {
-        if (err && err.message !== 'read ECONNRESET') {
```
richardschneider: Why the removal.  Normally `ipfs.stop` should not return an error.  There is a bug in go, in that it closes the connection before sending the result.

js/src/miscellaneous.js
```diff
@@ -83,12 +83,11 @@ module.exports = (common) => {
...
-          expect(err).to.not.exist()
-        }
+        // Retry
         ipfs.stop((err) => {
           expect(err).to.exist()
```
richardschneider: error should not exist.

js/src/miscellaneous.js
vmx: The reason is that the error message on Browsers is different from `read ECONNRESET`. So this check would fail, and the whole test case would fail. Hence I allow one error (no matter which message) to happen (due to the bug in Go).

js/src/miscellaneous.js
richardschneider: Gotcha, then we need to check for `go` and then expect an error

```js
if (!withGo)
   expect(err).to.not.exist()
```

Here's an [example](https://github.com/ipfs/interface-ipfs-core/blob/c4934ca0b3b43f5bfc1ff5dd38f85d945d3244de/js/src/stats.js#L23) of detecting `go-ipfs`.

vmx:
I finally understood what the test is actually doing. I added a commit which hopefully also makes it clearer for others.

diasdavid:
@vmx do you confirm that this indeed passes with js-ipfs and js-ipfs-api? If so, I'll merge and release

vmx:
@diasdavid I confirm that the miscellaneous.spec.js passes with js-ipfs-api on Windows and Linux. js-ipfs doesn't run these tests at all atm.

diasdavid:
@vmx does run them as well, it just never got the file renamed, still named generic.js https://github.com/ipfs/js-ipfs/blob/master/test/core/interface/generic.js

vmx:
@diasdavid js-ipfs generic.js passes on Windows and Linux.

PR #236, opened 2018-03-13
fix: fix broken stats test

vmx:
One test broke due to c4934ca0b3b43f5bfc1ff5dd38f85d945d3244de [1].

If promises are used, you needn't pass `done` as parameter of
the callback in.

[1]: https://github.com/ipfs/interface-ipfs-core/commit/c4934ca0b3b43f5bfc1ff5dd38f85d945d3244de#diff-0a6449ecfa8b9e3d807f53dde24eca71R66

### DIFF
--- a/js/src/files.js
+++ b/js/src/files.js
@@ -919,8 +919,8 @@ module.exports = (common) => {
       })
     })
 
-    describe('.lsPullStream', function (done) {
-      before((done) => {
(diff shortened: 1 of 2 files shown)


vmx:
@MichaelMure See this PR for future reference. I also wasn't aware that passing in the `done` argument could be an issue.

Issue #226, opened 2018-02-28
file.content is a stream and toString does not work


kevinsimper:
Is it a mistake that it tries to do a `file.content.toString()` when the file.content is a stream.

https://github.com/ipfs/interface-ipfs-core/blame/master/SPEC/FILES.md#L333

![image](https://user-images.githubusercontent.com/1126497/36788240-6b77e0a8-1c8d-11e8-83af-f54b82ecb854.png)

I can easily fix it if we agree :)

diasdavid:
@kevinsimper the line you pointed to is doing `file.path.toString()`, not _content_.

kevinsimper:
@diasdavid Yes, that is probrably another mistake as the comment says above `
// write the file's path and contents to standard out`
https://github.com/ipfs/interface-ipfs-core/blame/master/SPEC/FILES.md#L331

diasdavid:
@kevinsimper if you want to get a buffer, use `ipfs.files.get` instead

diasdavid:
Thanks for fixing with #237 <3

PR #237, opened 2018-03-14
Update with example outputting the content to stdout

kevinsimper:
I am updating with a example that does what the comment says `write the file's path and contents to standard out`. It takes the getting started docs and outputs them to the console.

https://ipfs.io/docs/getting-started/

https://jsbin.com/niwadeceze/1/edit?js,output

Closing https://github.com/ipfs/interface-ipfs-core/issues/226

### DIFF
--- a/SPEC/FILES.md
+++ b/SPEC/FILES.md
@@ -323,14 +323,19 @@ It returns a [Readable Stream][rs] in [Object mode](https://nodejs.org/api/strea
 **Example:**
 
 ```JavaScript
-const validCID = 'QmQ2r6iMNpky5f1m4cnm3Yqw8VSvjuKpTcK1X7dBR1LkJF'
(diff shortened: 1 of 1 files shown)


SPEC/FILES.md
```diff
@@ -323,14 +323,19 @@ It returns a [Readable Stream][rs] in [Object mode](https://nodejs.org/api/strea
...
 
 stream.on('data', (file) => {
   // write the file's path and contents to standard out
   console.log(file.path)
-  console.log(file.path.toString())
+  if(file.type !== 'dir') {
+    file.content.on('data', (data) => {
+      console.log(data.toString())
+    })
+    file.content.resume()
```
diasdavid: Does it require the `resume` call?

kevinsimper: Yes I tried without and it did not work. You can see the proof in the jsbin link :)

diasdavid: @kevinsimper thanks for checking. I've seen this in the past too. Let's keep it in the docs then :)

PR #241, opened 2018-03-21
Js ipfs issue 1145

youngnicks:
Adds tests for cat errors when dealing with nested links. If `No such file` is still needed as a valid value, I can add it back.

### DIFF
--- a/js/src/files.js
+++ b/js/src/files.js
@@ -441,10 +441,29 @@ module.exports = (common) => {
         return ipfs.files.cat(smallFile.cid + '/does-not-exist')
           .catch((err) => {
             expect(err).to.exist()
-            expect(err.message).to.contain.oneOf([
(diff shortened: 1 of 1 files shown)


youngnicks:
I did not try it with `js-ipfs-api`, but I will after I get the final error messages sorted in 1145.

youngnicks:
I'm having some issues getting `js-ipfs-api` to use the new code. I installed my local versions of `js-ipfs` and `interface-ipfs-core`, but it seems to still be giving a different error. Do you guys have any tips on testing `js-ipfs-api`?

youngnicks:
Added go-ipfs version of error messages. From reading the docs, it sounds like it can connect to a go-ipfs node, which would give different error messages until the go-ipfs error messages are updated to match js-ipfs cat error messages.

PR #242, opened 2018-03-21
doc: Documents the offset/length arguments to ipfs.files.cat and friends

achingbrain:
Requires ipfs/js-ipfs-unixfs-engine#207 and ipfs/js-ipfs#1231 to be merged first.

### DIFF
--- a/SPEC/FILES.md
+++ b/SPEC/FILES.md
@@ -159,9 +159,9 @@ pull(
 
 ##### `Go` **WIP**
 
-##### `JavaScript` - ipfs.files.cat(ipfsPath, [callback])
+##### `JavaScript` - ipfs.files.cat(ipfsPath, [options], [callback])
 
-ipfsPath can be of type:
+`ipfsPath` can be of type:
 
 - [`cid`][cid] of type:
   - a [CID](https://github.com/ipfs/js-cid) instance
@@ -171,6 +171,10 @@ ipfsPath can be of type:
   - '/ipfs/QmXEmhrMpbVvTh61FNAxP9nU7ygVtyvZA8HZDUaqQCAb66'
   - '/ipfs/QmXEmhrMpbVvTh61FNAxP9nU7ygVtyvZA8HZDUaqQCAb66/a.txt'
   - 'QmXEmhrMpbVvTh61FNAxP9nU7ygVtyvZA8HZDUaqQCAb66/a.txt'
+
+`options` is an optional object that may contain the following keys:
+  - `offset` is an optional byte offset to start the stream at
+  - `length` is an optional number of bytes to read from the stream
 
 `callback` must follow `function (err, file) {}` signature, where `err` is an error if the operation was not successful and `file` is a [Buffer][b]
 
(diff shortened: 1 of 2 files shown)


SPEC/FILES.md
```diff
@@ -161,9 +161,9 @@ pull(
 
 ##### `Go` **WIP**
 
-##### `JavaScript` - ipfs.files.cat(ipfsPath, [callback])
+##### `JavaScript` - ipfs.files.cat(ipfsPath, [begin], [end], [callback])
```
diasdavid: Why args like this? Let's add `options` and have that begin and end there.

SPEC/FILES.md
```diff
@@ -197,9 +199,9 @@ A great source of [examples][] can be found in the tests for this API.
 
 ##### `Go` **WIP**
 
-##### `JavaScript` - ipfs.files.catReadableStream(ipfsPath) -> [Readable Stream][rs]
+##### `JavaScript` - ipfs.files.catReadableStream(ipfsPath, [begin], [end]) -> [Readable Stream][rs]
```
diasdavid: Same

SPEC/FILES.md
```diff
@@ -225,9 +230,9 @@ A great source of [examples][] can be found in the tests for this API.
 
 ##### `Go` **WIP**
 
-##### `JavaScript` - ipfs.files.catPullStream(ipfsPath) -> [Pull Stream][rs]
+##### `JavaScript` - ipfs.files.catPullStream(ipfsPath, [begin], [end]) -> [Pull Stream][rs]
```
diasdavid: Same

SPEC/FILES.md
achingbrain: Fair enough, will change.  I've noticed that we have a [`ipfs.files.read`](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#read) function that accepts `offset` and `count` - it might be better to use that instead of `begin` and `end` to be consistent.  What do you think?

Also, I can't find an implementation of that in js-ipfs but it looks like it could just call through to `ipfs.files.cat`?

diasdavid: >  offset and count

👍 

> ipfs.files.read

This is a MFS specific operation.

achingbrain: > > offset and count
>
> 👍

Done

richardschneider:
The go documentation uses `offset` and `length`.  Should we use these? `offset` and `begin` are the same except for naming; whereas `length` and `end` have different semantics.

achingbrain:
@diasdavid I was going to move some of the tests from `js-ipfs-unixfs-engine` to this module as a follow up in order to not break anything if the modules were released one at a time.  Happy to do it all in one go if we're manually batching up module releases.

@richardschneider Yes, we'd discussed that on the outdated changes that Github has hidden above.  The currently go-only [`ipfs.files.read`](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/FILES.md#read) function uses `offset` and `count` instead of `offset` and `length` however which reads a little weirdly to me but agreed consistency is more important.

achingbrain:
After a straw poll in the js-dev catchup it was decided that `offset/length` is preferable to `offset/count`.

@diasdavid I've added some tests for stream slices that should only run for js-ipfs.

There's quite a lot of scope for test duplication across all these modules, what's the general approach?  In depth tests here and 'smoke test' style tests in dependencies?

PR #248, opened 2018-03-29
[WIP] adds tests for name.publish and name.resolve

alanshaw:
* Adds test which:
    1. creates two nodes
    2. connects them (`swarm.connect`)
    3. calls `name.publish` on node0
    4. calls `name.resolve(`/ipns/${node0.peerid}`)` on the other

**I could do with some advice possibly from the go-ipfs team(?), as step 4 fails with `Error: Could not resolve name.`. Why might that be?**

Also:

* Pulls repeated `spawnWithId` function into a module
* Updates doc to reflect return value from go-ipfs


VictorBjelkholm:
Should try to use the new experimental feature for pubsub resolve (but still test the old one as well), but with the flag `--enable-namesys-pubsub` together with `--enable-pubsub-experiment` will make publish/resolve a lot faster.

diasdavid:
What's missing here?

alanshaw:
Just the fact that the tests don't pass and I couldn't figure out why...

PR #252, opened 2018-04-04
feat: Provide access to bundled libraries when in browser

vasco-santos:
This PR aims to update the spec, as a result of the [Issue 406](https://github.com/ipfs/js-ipfs-api/issues/406).

According to the following PRs:

- `js-ipfs`: [PR 1297](https://github.com/ipfs/js-ipfs/pull/1297)
- `js-ipfs-api`:  [PR 732](https://github.com/ipfs/js-ipfs-api/pull/732)


SPEC/TYPES.md
```diff
@@ -0,0 +1,16 @@
...
+A set of data types are exposed directly from the IPFS instance under `ipfs.types`. That way you're not required to import/require the following.
+
+- [`ipfs.types.Buffer`](https://www.npmjs.com/package/buffer)
+- [`ipfs.types.PeerId`](https://github.com/libp2p/js-peer-id)
+- [`ipfs.types.PeerInfo`](https://github.com/libp2p/js-peer-info)
+- [`ipfs.types.multiaddr`](https://github.com/multiformats/js-multiaddr)
+- [`ipfs.types.multibase`](https://github.com/multiformats/multibase)
+- [`ipfs.types.multihash`](https://github.com/multiformats/js-multihash)
+- [`ipfs.types.CID`](https://github.com/ipld/js-cid)
+- [`ipfs.types.crypto`](https://github.com/libp2p/js-libp2p-crypto)
```
diasdavid: Not a type.

SPEC/TYPES.md
```diff
@@ -0,0 +1,16 @@
...
+- [`ipfs.types.dagPB`](https://github.com/ipld/js-ipld-dag-pb)
+- [`ipfs.types.dagCBOR`](https://github.com/ipld/js-ipld-dag-cbor)
+- [`ipfs.types.isIPFS`](https://github.com/ipfs-shipyard/is-ipfs)
```
diasdavid: Not a type.

SPEC/TYPES.md
```diff
@@ -0,0 +1,16 @@
...
+- [`ipfs.types.Buffer`](https://www.npmjs.com/package/buffer)
+- [`ipfs.types.PeerId`](https://github.com/libp2p/js-peer-id)
+- [`ipfs.types.PeerInfo`](https://github.com/libp2p/js-peer-info)
+- [`ipfs.types.multiaddr`](https://github.com/multiformats/js-multiaddr)
+- [`ipfs.types.multibase`](https://github.com/multiformats/multibase)
+- [`ipfs.types.multihash`](https://github.com/multiformats/js-multihash)
+- [`ipfs.types.CID`](https://github.com/ipld/js-cid)
+- [`ipfs.types.crypto`](https://github.com/libp2p/js-libp2p-crypto)
+- [`ipfs.types.dagPB`](https://github.com/ipld/js-ipld-dag-pb)
+- [`ipfs.types.dagCBOR`](https://github.com/ipld/js-ipld-dag-cbor)
```
diasdavid: perhaps scope these under IPLD and add all the ones that are supported today?

SPEC/TYPES.md
vasco-santos: Agreed. Should I create `util` scope for encompassing this?

SPEC/TYPES.md
vasco-santos: I agree with your idea. However, I would have to add `ipld/js-cid` too. As a consequence, a breaking change would be created in `js-ipfs` [core/index.js#L62](https://github.com/ipfs/js-ipfs/blob/master/src/core/index.js#L62). What do you think?

Regarding adding the other ones, if we decide for creating the specific scope, I think we may add them as well. So, you are referring to all the not archived JS modules at [ipld](https://github.com/ipld?utf8=%E2%9C%93&q=&type=&language=javascript)?

SPEC/TYPES.md
diasdavid: Good idea!

SPEC/TYPES.md
diasdavid: That's a valid point, thanks for thinking of that. Let's keep it as it is then.

vasco-santos:
Done @diasdavid 😄 

I also updated the other PRs accordingly.

js/src/util.js
```diff
@@ -0,0 +1,40 @@
...
+'use strict'
+
+const crypto = require('libp2p-crypto')
+const isIPFS = require('is-ipfs')
+
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const expect = chai.expect
+chai.use(dirtyChai)
+util
```
alanshaw: `undefined`?

js/src/types.js
```diff
@@ -0,0 +1,53 @@
...
+const multibase = require('multibase')
+const multihash = require('multihashes')
+const CID = require('cids')
+
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const expect = chai.expect
+chai.use(dirtyChai)
+
+describe('.types', function () {
```
alanshaw: Are these runnable? There's no `module.exports = (common) => {` wrapper and no entry in `js/src/index.js` for them.

js/src/util.js
vasco-santos: That util was supposed to be inside `describe`. How should I proceed @diasdavid ? A new PR fixing this typo?

js/src/types.js
diasdavid: @vasco-santos did you open a new PR?

js/src/util.js
diasdavid: @vasco-santos yes please.

js/src/types.js
vasco-santos: It is created now 🙂

PR #255, opened 2018-04-17
fix: skips the pin tests if js-ipfs detected

alanshaw:


vmx:
If I see it correctly all tests are skipped. Currently if none of the tests work on js-ipfs the tests are just not included in `https://github.com/ipfs/js-ipfs/tree/master/test/core` and hence not run. So this PR is not really needed.

alanshaw:
I closed this because I came to the conclusion that interface-ipfs-core shouldn't really be determining which tests should run on which implementations. i.e. If the tests don't pass, it's not compatible.

Just a note though that there is precedent here - _all_ the MFS tests are currently skipped for non-go implementations.

On that note, all those tests are skipped with `if (!withGo)` which means that those tests won't be run for ALL implementations that aren't go-ipfs whereas I think the intent is to not run them on js-ipfs.

PR #258, opened 2018-04-23
chore: Use normal functions so this.skip has the correct context

achingbrain:
Removes fat arrow functions in favour of `function` functions so `this.skip` can be used.

### DIFF
--- a/js/src/files.js
+++ b/js/src/files.js
@@ -559,7 +559,7 @@ module.exports = (common) => {
           })
       })
 
-      it('exports a chunk of a file', (done) => {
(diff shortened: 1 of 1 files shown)


diasdavid:
Didn't I fix this with https://github.com/ipfs/interface-ipfs-core/commit/2545ddd3dd6cdc46a3aef340cd26cf23460a4708? Why isn't it showing as a merge conflict, interesting.

achingbrain:
Weird, maybe it's because the changes are now identical to master?  Anyway, this is no longer required.

PR #263, opened 2018-05-05
Faster tests

VictorBjelkholm:
- Makes the pubsub "load tests" to act less like load tests, by just lowering the amount of messages. Load tests should be called outside of the normal tests instead.


VictorBjelkholm:
Alrighty, signed commits, proper commit message and sign-off in the commit message. Ready :)

PR #267, opened 2018-05-10
test: Bitswap wantlist peerid

wraithgar:
Tests for wantlist including with peerId param.

Still need to add copy to SPEC/BITSWAP.md

ETA branched from https://github.com/ipfs/interface-ipfs-core/pull/254 and that PR should merge before this one.


js/src/bitswap.js
```diff
@@ -0,0 +1,137 @@
...
+
+    it('.wantlist gives error if offline', () => {
+      ipfs.bitswap.wantlist((err, list) => {
+        expect(err).to.exist()
+        expect(list).to.not.exist()
+      })
+    })
+
+    it('.unwant gives error if offline', () => {
+      expect(() => ipfs.bitswap.unwant('my key'), (err) => {
```
diasdavid: This will always be an invalid key.

daviddias (changes requested):
Cool stuff! Did these tests got run against js-ipfs-api and js-ipfs?

wraithgar:
@diasdavid oh yeah looks like I forgot to push and make the PR for the js-ipfs-api code this requires.

js/src/bitswap.js
wraithgar: Ok changed that to use a proper key and still expect an error.

js/src/bitswap.js
```diff
@@ -0,0 +1,137 @@
...
+              const ipfsBAddr = id.addresses[0]
+              ipfsBId = id.id
+              ipfsA.swarm.connect(ipfsBAddr, cb)
+            })
+          },
+          (cb) => {
+            ipfsB.block.get(new CID(key))
+              .then(() => {})
+              .catch(() => {})
+            setTimeout(cb, 500)
```
alanshaw: Could you explain what's going on here? Maybe also add a comment?

js/src/bitswap.js
```diff
@@ -0,0 +1,137 @@
...
+        expect(err).to.not.exist()
+        expect(list[0].cid.toBaseEncodedString()).to.equal(key)
+        done()
+      })
+    })
+
+    it('.unwant', (done) => {
+      ipfsA.bitswap.unwant(new CID(key), (err) => {
+        ipfsA.bitswap.wantlist((err, list) => {
+          expect(list).to.be.empty()
```
alanshaw: Also `expect(err).to.not.exist()`?

js/src/bitswap.js
```diff
@@ -0,0 +1,137 @@
...
+        ], done)
+      })
+    })
+
+    after((done) => common.teardown(done))
+
+    it('.stat', (done) => {
+
+      ipfsA.bitswap.stat((err, stats) => {
+        statsTests.expectIsBitswap(err, stats)
```
alanshaw: Also `expect(err).to.not.exist()`?

js/src/bitswap.js
```diff
@@ -0,0 +1,137 @@
...
+              done()
+            })
+          })
+        })
+      })
+    })
+
+    it('.stat gives error while offline', () => {
+      ipfs.bitswap.stat((err, stats) => {
+        expect(err).to.exist()
```
alanshaw: Can we assert against the error message? Just expecting it to exist could give to false positives.

alanshaw (changes requested):
Thank you! Looks great, just a few minor points to address and it'll be good to merge 🚀

SPEC/BITSWAP.md
```diff
@@ -4,13 +4,37 @@
...
+`cid` is a [cid][cid] which can be passed as:
+
+- CID, a CID instance
+- String, the base58 encoded version of the multihash
+
+##### `Go` **WIP**
+
+### `bitswap.wantlist`
+
+> Returns the wantlist, optionally limited by peerID
```
alanshaw: It would be helpful to clarify if "the wantlist" when not limited by peer ID is a list that everybody shares or if it is just yours.

Also, would "filtered" be a more accurate description than "limited"?

SPEC/BITSWAP.md
```diff
@@ -4,13 +4,37 @@
...
+
+#### `Go` **WIP**
+
+#### `JavaScript` - ipfs.bitswap.wantlist([peerId])
+
+```JavaScript
+ipfs.bitswap.wantlist((err, list) => console.log(list))
+
+//[ { Wantlist object }, ... ]
```
alanshaw: Can you elaborate on the shape of the "Wantlist object" here? I think we usually try to in these specs.

alanshaw (changes requested):
Apologies, I noticed a couple more small issues after your changes.

alanshaw (approved):

PR #278, opened 2018-05-18
chore: enable-js-mfs-tests

achingbrain:


travisperson:
I think this is good to go now.

achingbrain:
No longer needed as we can enable/disable tests from the calling code.  🎉

PR #288, opened 2018-05-30
test: augment dht tests (first pass)

diasdavid:

PR #290, opened 2018-06-02
feat: modularise tests by command, add tools to skip and only

alanshaw:
This PR modularises tests by command and provides tools for consumers of this library to use skip/only on a command basis.

* Encourages more test isolation. Many tests were expecting values from previous tests to be available, this has largely been eradicated
* Allows us to test partially implemented sub-systems e.g. currently `repo.version` and `repo.stat` are implemented but `repo.gc` is not. This allows us to import tests for `repo.version` and `repo.stat` without failing the suite but also _import and skip_ tests for `repo.gc` for visibility until the implementation is complete
* Removes the need for implementation specific checks to skip tests i.e. `withGo` goes away because we can just run the tests we want to run on any given implementation
* Removes the need for platform specific checks to skip tests i.e. `isNode` goes away because we can just run the tests we want to run on any given platform
* Consistently expresses the test titles more descriptively as "it should..." or "it should not..." to allow for easier understanding of test purpose
* Consistently labels promised versions of tests with the same name as the non promised version but with the suffix "(promised)". Also moves these tests so they are adjacent in the suite file. This allows the developer to more easily see if the issue is because of the callback/promise interface for a command or a bug within the code for the command

~~This work is backwards compatible with the previous API so no changes to dependant code needs to be made.~~ I wanted this to be backwards compatible, but quickly ran into an issue where the `common` object we pass to suites expects to only be used once. I started to pass it to multiple suites and the `teardown` function was trying to tear down nodes that were already stopped! So, unfortunately **breaking change**: `common` should now be a function that creates a `common` object for use with the tests. I'll submit PR's to js-ipfs (https://github.com/ipfs/js-ipfs/pull/1389) and js-ipfs-api (https://github.com/ipfs/js-ipfs-api/pull/785) to deal with this.

**running tests by command**

```js
const tests = require('interface-ipfs-core')

tests.repo.version(common)
```

N.B. you can still run a whole sub-system suite like this:

```js
const tests = require('interface-ipfs-core')
tests.repo(common)
```

**skipping tests for a specific command**

```js
const tests = require('interface-ipfs-core')

tests.repo.version(common)
tests.repo.stat(common)
tests.repo.gc(common, { skip: true }) // pass an options object to skip these tests

// OR, at the subsystem level

tests.repo(common, { skip: ['gc'] })
```

**skipping specific tests**

```js
const tests = require('interface-ipfs-core')

tests.repo.version(common)
tests.repo.stat(common)
tests.repo.gc(common, { skip: ['should do a thing'] }) // named test(s) to skip

// OR, at the subsystem level

tests.repo(common, { skip: ['should do a thing'] })
```

---

**only**

The idea here is that `only` can be used during development/bug fixing to run only problematic tests without having to `npm link` and find the test(s) to add `.only` to.

**running only tests for a specific command**

```js
const tests = require('interface-ipfs-core')

tests.repo.version(common)
tests.repo.stat(common)
tests.repo.gc(common, { only: true }) // pass an options object to run only these tests

// OR, at the subsystem level

tests.repo(common, { only: ['gc'] })
```

**running only specific tests**

```js
const tests = require('interface-ipfs-core')

tests.repo.version(common)
tests.repo.stat(common)
tests.repo.gc(common, { only: ['should do a thing'] }) // only run these named test(s)

// OR, at the subsystem level

tests.repo(common, { only: ['should do a thing'] })
```


travisperson:
This hits a lot of points that Victor and I wanted to try and look into this quarter. I really like the way this is laid out and the removal of the `withGo` and other opt out variables.

I would love to help take this through to competition.

alanshaw:
@diasdavid @travisperson 

I appreciate this is a massive PR to review, but a lot of it is mechanical changes. **Please read the PR description above first** (a lot of this is echoed in the changes to the README). Just to highlight how things have moved around:

Each sub-system (dag, files, swarm etc.) is now a **folder** with an `index.js` file. It defines all tests available in that subsystem - so for `files`, the commands are `add`, `cat`, `get` etc..

All the tests for `files.add` are grouped in `src/files/add.js`, so `src/files/index.js` creates an object that includes all the tests for each command and calls a function called `createSuite`.

`createSuite` turns that object into a function that runs all of the tests (it looks like `function (createCommon, options) {}`) and **also** adds each command as a property. What's fun is that the function that `createSuite` returns (what you get back from `src/files/index.js`) looks just like the function you get from `src/files/add.js` so you can use them interchangeably to run ALL tests for `files` (by calling `tests.files()`) or just tests for `add` (by calling `tests.files.add()`).

---

BREAKING CHANGE: Consumers of this test suite now have fine grained control over what tests are run. Tests can now be skipped and "onlyed" (run only specific tests). This can be done on a test, command and sub-system level. See the updated usage guide for instructions.

This means that tests skips depending on implementation (e.g. go/js), environment (e.g. node/browser) or platform (e.g. macOS/linux/windows) that were previously present in this suite have been removed. Consumers of this library should add their own skips based on the implementation that's being tested and the environment/platform that the tests are running on.

The following other breaking changes have been made:

1. The common object passed to test suites has changed. It must now be a function that returns a common object (same shape and functions as before).
2. The `ipfs.ls` tests (not MFS `ipfs.files.ls`) is now a root level suite. You'll need to import it and use like `tests.ls(createCommon)` to have those tests run.
3. The `generic` suite (an alias to `miscellaneous`) has been removed.

js/src/dht/provide.js
```diff
@@ -0,0 +1,102 @@
...
+    after((done) => common.teardown(done))
+
+    it('should provide local CID', (done) => {
+      ipfs.files.add(Buffer.from('test'), (err, res) => {
+        if (err) return done(err)
+
+        ipfs.dht.provide(new CID(res[0].hash), (err) => {
+          expect(err).to.not.exist()
+          done()
+        })
```
travisperson: I don't believe many of these tests will work anymore as there will never be any peers on the dht to provide too, and there isn't much of a reason to provide to `self`.

The dht for go-ipfs I know will error if no peers are connected when running `provides` [0]. Should this be valid, and an issue opened for go-ipfs?



[0] https://github.com/ipfs/go-ipfs/blob/08fb11fa896704d9e081bf64e66fac1f6d9d03dc/core/commands/dht.go#L280

js/src/swarm/peers.js
```diff
@@ -0,0 +1,204 @@
...
+
+const expect = chai.expect
+chai.use(dirtyChai)
+
+module.exports = (createCommon, options) => {
+  const describe = getDescribe(options)
+  const it = getIt(options)
+  const common = createCommon()
+
+  describe('.swarm', function () {
```
travisperson: I believe this is `.swarm.peers` right?

js/src/swarm/local-addrs.js
```diff
@@ -0,0 +1,54 @@
...
+      this.timeout(100 * 1000)
+
+      common.setup((err, factory) => {
+        expect(err).to.not.exist()
+
+        spawnNodes(2, factory, (err, nodes) => {
+          expect(err).to.not.exist()
+          ipfs = nodes[0]
+          done()
+        })
```
travisperson: Have an extra node

js/src/swarm/addrs.js
```diff
@@ -0,0 +1,56 @@
...
+      // timeout for the before step
+      this.timeout(100 * 1000)
+
+      common.setup((err, factory) => {
+        expect(err).to.not.exist()
+
+        spawnNodes(2, factory, (err, nodes) => {
+          expect(err).to.not.exist()
+          ipfs = nodes[0]
+          done()
```
travisperson: Have an extra node?

js/src/dht/provide.js
alanshaw: I’ve run these tests against js-ipfs and js-ipfs-api (talks to go-ipfs) and they all pass, so the tests that could fail might already be skipped...I’ll check and get back to you

js/src/dag/put.js
```diff
@@ -0,0 +1,125 @@
...
+
+    it('should put dag-pb with custom hash func (sha3-512)', (done) => {
+      ipfs.dag.put(pbNode, {
+        format: 'dag-pb',
+        hashAlg: 'sha3-512'
+      }, done)
+    })
+
+    // This works because dag-cbor will just treat pbNode as a regular object
+    it.skip('should not put dag-pb node with wrong multicodec', (done) => {
```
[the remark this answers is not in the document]
alanshaw: Possibly, but it is currently skipped:

https://github.com/ipfs/interface-ipfs-core/blob/510516aaf5b310a15f413622352821f1e71bad0e/js/src/dag.js#L74-L80

js/src/dht/provide.js
alanshaw: DHT tests are completely commented out in js-ipfs currently 😟 https://github.com/ipfs/js-ipfs/blob/master/test/core/interface/dht.js

Some tests in js-ipfs-api were [already skipped](https://github.com/ipfs/js-ipfs-api/pull/785/files#diff-fd1dd60e18a7ade9d795b77bda4aff2eR47), but not any `provide` tests. I will investigate further...

alanshaw:
Sorry for the late adding of reviewers, I realised that this is a change that'll affect you all so I'd like you to all see it and know what is going on 🚀

js/src/block/get.js
```diff
@@ -0,0 +1,65 @@
+/* eslint-env mocha */
+'use strict'
+
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const multihash = require('multihashes')
+const CID = require('cids')
+const Buffer = require('safe-buffer').Buffer
```
diasdavid: No need for safe-buffer anymore.

js/src/bootstrap/rm.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-env mocha */
+'use strict'
+
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const { getDescribe, getIt } = require('../utils/mocha')
+
+const expect = chai.expect
+chai.use(dirtyChai)
```
diasdavid: If we are going to get `describe` and `it` through a local module, better to also pick `expect` from it, saving 4 lines from each module (and avoiding forgetting about it in the future).

js/src/bootstrap/rm.js
```diff
@@ -0,0 +1,76 @@
...
+
+const invalidArg = 'this/Is/So/Invalid/'
```
diasdavid: Move it inside the module.exports

js/src/bootstrap/rm.js
```diff
@@ -0,0 +1,76 @@
...
+
+module.exports = (createCommon, options) => {
+  const describe = getDescribe(options)
+  const it = getIt(options)
+  const common = createCommon()
```
diasdavid: Why does common need to created here, if there are no options for it?

js/src/dag/put.js
```diff
@@ -0,0 +1,125 @@
...
+    })
+
+    it('should put dag-cbor with custom hash func (sha3-512)', (done) => {
+      ipfs.dag.put(cborNode, {
+        format: 'dag-cbor',
+        hashAlg: 'sha3-512'
+      }, done)
+    })
+
+    it('should not put dag-cbor node with wrong multicodec', function (done) {
```
diasdavid: No need to use function if no special timeout needs to be added.

js/src/dag/tree.js
```diff
@@ -0,0 +1,141 @@
+/* eslint-env mocha */
+'use strict'
+
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const { series, eachSeries } = require('async')
```
diasdavid: We use require('async/series') as a good practice to avoid requiring the whole module into our codebase.

js/src/dht/findpeer.js
```diff
@@ -0,0 +1,61 @@
...
+    it('should find other peers', (done) => {
+      nodeA.dht.findpeer(nodeB.peerId.id, (err, peer) => {
+        expect(err).to.not.exist()
+        // TODO upgrade the answer, format is weird
+        expect(peer[0].Responses[0].ID).to.be.equal(nodeB.peerId.id)
+        done()
+      })
+    })
+
+    it('should fail to find other peer if peer does not exist', function (done) {
```
diasdavid: If no special timeout, please use a arrow func

js/src/dht/findpeer.js
```diff
@@ -0,0 +1,61 @@
...
+      nodeA.dht.findpeer('Qmd7qZS4T7xXtsNFdRoK1trfMs5zU94EpokQ9WFtxdPxsZ', (err, peer) => {
+        expect(err).to.not.exist()
+        expect(peer).to.be.equal(null)
```
diasdavid: nitpick `s/to.be.equal(null)/to.not.exist()`

js/src/dht/findprovs.js
```diff
@@ -0,0 +1,62 @@
...
+const expect = chai.expect
+chai.use(dirtyChai)
+
+module.exports = (createCommon, options) => {
+  const describe = getDescribe(options)
+  const it = getIt(options)
+  const common = createCommon()
+
+  describe('.dht.findprovs', function () {
+    this.timeout(80 * 1000)
```
diasdavid: Let's remove the global timeout in favor of by test timeout.

vasco-santos:
It seems a really better approach! Thanks @alanshaw

js/src/bootstrap/rm.js
alanshaw: We need to create a `common` object for each command now that they have been split into modules since it tracks nodes that are created so that they can be destroyed when teardown is called.

```js
// Before:
tests.files(common)
// -> calls:
tests.files.add(common) // teardown destroys nodes created for files.add
tests.files.get(common) // teardown destroys nodes created for files.add AND files.get - ruh roh!
// etc.

// Now:
tests.files(createCommon)
// -> calls:
tests.files.add(createCommon) // teardown destroys nodes created for files.add
tests.files.get(createCommon) // teardown destroys nodes created for files.get
```

The alternative would be to make the `common` object smarter - to remove nodes from it's `nodes` array when teardown is called. However I was thinking that if we were to ever run these suites concurrently (please! one day!) then things would break.

From the PR description:

> ~~This work is backwards compatible with the previous API so no changes to dependent code needs to be made.~~ I wanted this to be backwards compatible, but quickly ran into an issue where the `common` object we pass to suites expects to only be used once. I started to pass it to multiple suites and the `teardown` function was trying to tear down nodes that were already stopped! So, unfortunately **breaking change**: `common` should now be a function that creates a `common` object for use with the tests. I'll submit PR's to js-ipfs (https://github.com/ipfs/js-ipfs/pull/1389) and js-ipfs-api (https://github.com/ipfs/js-ipfs-api/pull/785) to deal with this.

js/src/dht/provide.js
alanshaw: These tests are passing for me for js-ipfs-api:

screen shot 2018-06-25 at 11 31 32

Is it because they are providing to the bootstrap nodes?

diasdavid: Historically there were only a couple of DHT tests. The suite was increased with https://github.com/ipfs/interface-ipfs-core/pull/288 which led me to found some issues with the http-api.

js/src/pubsub/peers.js
```diff
@@ -0,0 +1,151 @@
...
+      const ipfs3Addr = ipfs3.peerId.addresses.find((a) => a.includes('127.0.0.1'))
+
+      parallel([
+        (cb) => ipfs1.swarm.connect(ipfs2Addr, cb),
+        (cb) => ipfs1.swarm.connect(ipfs3Addr, cb),
+        (cb) => ipfs2.swarm.connect(ipfs3Addr, cb)
+      ], done)
+    })
+
+    describe('.peers', () => {
```
travisperson: Did you want to drop this describe as well?

alanshaw:
I think I've addressed all the feedback now, if there's no further comments then I'll merge this in tomorrow. Thank you and much ❤️ for your time and thoughts.

PR #297, opened 2018-06-08
Add docs for the pin option for files.add*

### LINKED ISSUE #294: Mention that files.add has a pin option
To quote lidel's code in irc:
`
ipfs.files.add(Buffer.from('auto-pin')).then(results => ipfs.pin.ls(results[0].hash).then(pins => console.dir(pins))).catch(console.error)
`
vs.
`
ipfs.files.add(Buffer.from('no-auto-pin'), {pin: false}).then(results => ipfs.pin.ls(results[0].hash).then(pins => console.dir(pins))).catch(console.error)
`

Should also mention that it automatically pins. (Do I have to add recursively to get it to pin recursively?)

Powersource:
Took the description from go-ipfs.

Fixes  #294


alanshaw:
Many thanks ❤️, would you be willing to send another PR adding tests for this feature?

PR #302, opened 2018-06-14
test: adds test vulnerable to encoding translation

achingbrain:
This follows on from ipfs/js-ipfs#1398 and tests the more general case.

Seems to fail with the current master of js-ipfs-api.


achingbrain:
This requires go-ipfs 0.4.17 to be released first - not much point merging this until then.

PR #308, opened 2018-06-21
test: add test for retrieving an empty block

alanshaw:
Add an empty block and ensure we can get it back successfully. Test for https://github.com/ipfs/js-ipfs-api/pull/789


alanshaw:
Yes, perhaps as a separate test in `block.put`?

travisperson:
Ah yes, that can come in later. Missed that this was a `get` test.

PR #311, opened 2018-06-26
fix: do not rely on discovery for ping tests

### LINKED ISSUE #310: util: waitUntilConnected
I've been going through the code and identifying patterns or ways to help clean up the code to make things easier to read and understand.

(Note: I've been working off of the in-progress PR #290)

There are a few tests that require multiple nodes to be connected (ping, pubsub, dht).

Running through this I found this setup code for `ping`, which uses a method [`waitUntilConnected`][0], which doesn't actually connect any nodes, but instead just waits around till the nodes find each other (as far as I can tell).

https://github.com/ipfs/interface-ipfs-core/blob/f238e9f0ba8132dba51d3bac1eb491701f505ccd/js/src/ping.js#L38-L52

It appears that it heavily relies on peer discovery, which is dependent on how the node is configured, particularity for js-ipfs-core.

@alanshaw I believe you wrote this code. I believe it was added so that the tests itself did not have to attempt to find and connect to the peers, and instead could rely on the peers already being connected before conducting the pings.

Any reasons to use `waitUntilConnected` over simply calling `ipfs.swarm.connect`?

[0]: https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/utils/connections.js#L3-L28

alanshaw:
The ping tests do not need to test that peers can discover each other. This PR removes the `waitUntilConnected` utility and explicitly connects the nodes that are used in the tests.

fixes #310


js/src/utils/spawn.js
```diff
@@ -24,7 +26,10 @@ exports.spawnNodes = spawnNodes
 
 // Spawn n nodes, getting their id's and setting them as `peerId` on the nodes
 function spawnNodesWithId (n, factory, callback) {
-  timesSeries(n, (_, cb) => spawnNodeWithId(factory, cb), callback)
+  spawnNodes(n, factory, (err, nodes) => {
+    if (err) return callback(err)
+    map(nodes, identify, callback)
+  })
```
alanshaw: Apologies the change in this file should have been proposed a separate PR - this simply ensures the ID calls are done in parallel which will speed up the tests a little.

travisperson:
LGTM!

PR #315, opened 2018-06-29
test: backport of ipfs/interface-ipfs-core#314

achingbrain:
See ipfs/interface-ipfs-core#314 for context

### DIFF
--- a/js/src/dag.js
+++ b/js/src/dag.js
@@ -71,14 +71,6 @@ module.exports = (common) => {
         }, done)
       })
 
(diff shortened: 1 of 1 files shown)


achingbrain:
This exists so we can get mfs in ahead of landing the various PRs that support ipfs/interface-ipfs-core#290

achingbrain:
> Oh, wow, I didn't know that there are maintenance branches now.

I don't think there are, as a rule, but ipfs/interface-ipfs-core#290 still has a bunch of work that needs doing on other repos so this is a stopgap to keep the release train rolling.

PR #316, opened 2018-06-29
test(dag): ensure `dag.put` can be called without options

PascalPrecht:
As discussed in https://github.com/ipfs/js-ipfs/pull/1415#issue-198357338


js/src/dag/put.js
```diff
@@ -113,6 +113,10 @@ module.exports = (createCommon, options) => {
       })
     })
 
+    it('shouldn\'t fail when calling put without options', (done) => {
```
alanshaw: For consistency could you please use "should not" over "shouldn't"

PascalPrecht: Done 👍

PascalPrecht:
@alanshaw updated the spec document. Let me know if that's how you imagined it, or if the wording needs change.

PR #324, opened 2018-07-05
chore: re-enable linting

alanshaw:
When the JS code was moved into the js/ directory the aegir linting task was no longer considering them for linting. This is because the default globs do not include this path.

This PR adds custom lint globs to the aegir config and fixes linting errors.

depends on ipfs/aegir#240

License: MIT
Signed-off-by: Alan Shaw <alan@tableflip.io>

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,0 +1,10 @@
+'use strict'
+
+module.exports = {
+  lint: {
+    files: [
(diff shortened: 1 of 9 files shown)


alanshaw:
I picked off most of them when I did the modularization PR 😉

PR #326, opened 2018-07-16
[WIP] fix: unsubscribe in series for go-ipfs

alanshaw:
Towards js-ipfs-api perma-green

### DIFF
--- a/js/src/pubsub/unsubscribe.js
+++ b/js/src/pubsub/unsubscribe.js
@@ -1,7 +1,7 @@
 /* eslint-env mocha */
 'use strict'
 
-const each = require('async/each')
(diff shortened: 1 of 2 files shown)


js/src/pubsub/unsubscribe.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-env mocha */
 'use strict'
 
-const each = require('async/each')
+const eachSeries = require('async/each')
```
VictorBjelkholm: This is actually using `async/each` here and not `eachSeries`. There is a typo on one side :)

alanshaw: ha thanks

PR #327, opened 2018-07-16
feat: ipns working locally

vasco-santos:
PR in the context of [1400 - IPNS working locally](https://github.com/ipfs/js-ipfs/pull/1400)


js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+      this.timeout(60 * 1000)
+
+      common.setup((err, factory) => {
+        expect(err).to.not.exist()
+
+        factory.spawnNode((err, node) => {
+          expect(err).to.not.exist()
+
+          ipfs = node
+          ipfs.id().then((res) => {
```
alanshaw: Please use [`spawnNodeWithId`](https://github.com/ipfs/interface-ipfs-core/blob/34c3fc65f105528dfe17b2c9cc5ee50bea113e60/js/src/utils/spawn.js#L16) to simplify this!

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+      function populate () {
+        each(fixtures.files, (file, cb) => {
+          ipfs.files.add(file.data, { pin: false }, cb)
+        }, done)
+      }
+    })
+
+    after((done) => common.teardown(done))
+
+    it('name publish should publish correctly', (done) => {
```
alanshaw: "name publish should publish correctly" => "should publish correctly"

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+      ipfs.name.publish(value, true, '1m', '10s', 'self', (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+        expect(res).to.equal(nodeId)
+
+        done()
+      })
+    })
+
+    it('name publish should publish correctly when the file was not added but resolve is disabled', (done) => {
```
alanshaw: "name publish should publish correctly when the file was not added but resolve is disabled" => "should publish correctly when the file was not added but resolve is disabled"

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+      ipfs.name.publish(value, false, '1m', '10s', 'self', (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+        expect(res).to.equal(nodeId)
+
+        done()
+      })
+    })
+
+    it('name publish should publish correctly when a new key is used', (done) => {
```
alanshaw: "name publish should publish correctly when a new key is used" => "should publish correctly when a new key is used"

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+        }, done)
+      }
+    })
+
+    after((done) => common.teardown(done))
+
+    it('name publish should publish correctly', (done) => {
+      const value = fixtures.files[0].cid
+
+      ipfs.name.publish(value, true, '1m', '10s', 'self', (err, res) => {
```
alanshaw: This doesn't follow the signature described in https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/NAME.md#namepublish - should the parameters other than `value` and the callback be in an options object?

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+const { fixtures } = require('./utils')
+const { getDescribe, getIt, expect } = require('../utils/mocha')
+
+module.exports = (createCommon, options) => {
+  const describe = getDescribe(options)
+  const it = getIt(options)
+  const common = createCommon()
+
+  describe('.name.publish', function () {
+    this.timeout(50 * 1000)
```
alanshaw: Traditionally we've favored test specific timeouts over global timeouts so that we're alerted if a single test that usually takes a very short amount of time starts to take a long amount of time. Can we please remove this and add increased timeouts to the tests that need them?

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+
+    after((done) => common.teardown(done))
+
+    it('name publish should publish correctly', (done) => {
+      const value = fixtures.files[0].cid
+
+      ipfs.name.publish(value, true, '1m', '10s', 'self', (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+        expect(res).to.equal(nodeId)
```
alanshaw: When you publish in go-ipfs you get back the mapping of name => value. This is also what the spec says here: https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/NAME.md#namepublish

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+
+        done()
+      })
+    })
+
+    it('name publish should publish correctly when a new key is used', (done) => {
+      const value = fixtures.files[0].cid
+
+      ipfs.name.publish(value, false, '24h', '10s', keyName, (err, res) => {
```
alanshaw: This is the only test that uses `keyName` in this suite, I'd move the key generation here.

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+            })
+          })
+        })
+      })
+
+      function populate () {
+        each(fixtures.files, (file, cb) => {
+          ipfs.files.add(file.data, { pin: false }, cb)
+        }, done)
+      }
```
alanshaw: Please move this into a separate `before` block (they are executed in series in the order they appear) so the setup is a bit easier to read.

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+          ipfs.id().then((res) => {
+            expect(res.id).to.exist()
+
+            nodeId = res.id
+
+            ipfs.key.gen(keyName, { type: 'rsa', size: 2048 }, (err, key) => {
+              expect(err).to.not.exist()
+              expect(key).to.exist()
+              expect(key).to.have.property('name', keyName)
+              expect(key).to.have.property('id')
```
alanshaw: We're not testing `ipfs.key.gen`, so there's no need to assert on the result.

js/src/name/publish.js
```diff
@@ -0,0 +1,96 @@
...
+              populate()
+            })
+          })
+        })
+      })
+
+      function populate () {
+        each(fixtures.files, (file, cb) => {
+          ipfs.files.add(file.data, { pin: false }, cb)
+        }, done)
```
alanshaw: It looks as though the tests only use one of these files, can we get some speed ups by only adding what we use?

js/src/name/resolve.js
```diff
@@ -0,0 +1,123 @@
...
+          })
+        })
+      })
+
+      function populate () {
+        each(fixtures.files, (file, cb) => {
+          ipfs.files.add(file.data, { pin: false }, cb)
+        }, done)
+      }
+    })
```
alanshaw: Please see comments for the before block in name.publish - I think they mostly all apply here too.

js/src/name/resolve.js
```diff
@@ -0,0 +1,123 @@
...
+
+    after((done) => common.teardown(done))
+
+    it('name resolve should resolve correctly after a publish', (done) => {
```
alanshaw: Please remove the "name resolve " prefix from all of these test names so that they start with "should" :)

js/src/name/resolve.js
```diff
@@ -0,0 +1,123 @@
...
+      this.timeout(60 * 1000)
+
+      const value = fixtures.files[0].cid
+
+      ipfs.name.publish(value, true, '1m', '10s', 'self', (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+
+        ipfs.name.resolve(nodeId, false, false, (err, res) => {
```
alanshaw: This signature doesn't follow the spec document https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/NAME.md#nameresolve can `false, false` be an options object instead? I think this is more readable.

js/src/name/resolve.js
```diff
@@ -0,0 +1,123 @@
...
+    it('name resolve should not get the entry correctly if its validity time expired', (done) => {
+      this.timeout(60 * 1000)
+
+      const value = fixtures.files[0].cid
+
+      ipfs.name.publish(value, true, '10ns', '10s', 'self', (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+
+        setTimeout(function () {
```
alanshaw: Is this required? There might be a bug if we can't immediately resolve a value after publishing it.

js/src/name/resolve.js
```diff
@@ -0,0 +1,123 @@
...
+            expect(err).to.exist()
+            expect(res).to.not.exist()
+
+            done()
+          })
+        }, 1)
+      })
+    })
+
+    it('name resolve should should go recursively until finding an ipfs hash', (done) => {
```
alanshaw: Could we re-word this one?

How about "should recursively resolve to an IPFS hash"?

alanshaw:
Also @vasco-santos for the future, I have an open PR for adding a (non-local) `name.*` test, which you might want to pull over the finish line if you implement non-local IPNS later ;) https://github.com/ipfs/interface-ipfs-core/pull/248/files

js/src/name/publish.js
VictorBjelkholm: Instead of writing "correctly" we should write what we're expecting with the test cases. "correctly" is up to the eye of the beholder :)

Also, if `should` is something we're adding before every title, we should just remove it. So this case ends up being "Publish correctly", which gives me the hint that the test is either trying to span too many things or is simply irrelevant.

VictorBjelkholm: Note: irrelevant as in the other tests will tell us exactly what "correctly" means, so might be enough with the rest of the tests.

js/src/name/resolve.js
VictorBjelkholm: Disagree, no need to add "should" before every test case, it's just mindless repetition for no gain.

vmx: I prefer having "should" as you can the really read it and check if it makes sense. I've seen too many test descriptions in my past lives where it wasn't really clear if it's the test or the error condition it's describing.

VictorBjelkholm: All the tests are defining what the test "should" do, that's the point of the tests. Now it's not a super big deal, so I'm fine with that if that's what we end up deciding. 

> where it wasn't really clear if it's the test or the error condition it's describing.

This is a separate issue though, and has nothing to do with if the test name is prefixed with "should" or not.

js/src/name/publish.js
vasco-santos: I didn't know that function! Thanks, I will use it :)

js/src/name/publish.js
vasco-santos: true! I will modify it

js/src/name/resolve.js
vasco-santos: I also prefer to have `should` prefixing the test description, as it seems more readable. But I can also remove it if we decide that.

js/src/name/resolve.js
vasco-santos: Yah, I agree. I did it only on the top level (CLI, HTTP), but I agree with having this as well here

js/src/name/resolve.js
vasco-santos: The `setTimeout` here aims to guarantee that the record has an expired validity. This way, when we try to resolve it, its validity has expired and we will receive that error, instead of the value.

vasco-santos:
> Also @vasco-santos for the future, I have an open PR for adding a (non-local) `name.*` test, which you might want to pull over the finish line if you implement non-local IPNS later ;) https://github.com/ipfs/interface-ipfs-core/pull/248/files

Cool @alanshaw . I will look at it when implementing the non-local IPNS!

js/src/name/publish.js
vasco-santos: I added some information, in order to be more understandable: 

`should publish an IPNS record with the default params`

I believe this way is better 😄

js/src/name/publish.js
vasco-santos: With the `before` refactor, and only adding a single file now, I think that it will be better to have all together. When I submit a new commit, let me know if you agree 🙂

js/src/name/publish.js
vasco-santos: I was only doing this on the top level and not here. I will change this 😄

vasco-santos: Two other things regarding those specs.

First, the docs [spec](https://ipfs.io/docs/api/#apiv0namepublish) for publish, have the `Name` and `Value` capitalized, while the `NAME.md` spec hasn't. Should we uniform this? Go uses capitalized as well.

Second, those SPECs for [resolve](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/NAME.md#nameresolve), they are not consistent with the docs [spec](https://ipfs.io/docs/api/#apiv0nameresolve).

Basically, the resolve should return an object with the `Path` as the docs spec have, instead of the `NAME.md` spec from this repo, right? I can also change that spec here.

vasco-santos:
@alanshaw your review was entirely addressed, and the PR for `js-ipfs` [#1400](https://github.com/ipfs/js-ipfs/pull/1400) was also updated accordingly.

Particularlly, I would like you to give a look [here](https://github.com/ipfs/interface-ipfs-core/pull/327#discussion_r204000963)

js/src/name/resolve.js
alanshaw: > All the tests are defining what the test "should" do

...or what they should not do 😉 

I think it's good practice to _encourage_ people to express tests in terms of what they should or should not do. I've seen tests here who's description was "it buffer" or "it big buffer" - that's not helpful to me when I look at the output. By having a convention of naming tests like "it should..." or "it should not..." I find that it encourages people to consider the rest of the sentence more carefully and we get better quality descriptions.

js/src/name/publish.js
```diff
@@ -0,0 +1,103 @@
...
+    it('should publish an IPNS record with the default params', (done) => {
+      this.timeout(50 * 1000)
+
+      const value = fixture.cid
+
+      ipfs.name.publish(value, (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+        expect(res.Name).to.equal(nodeId)
+        expect(res.Value).to.equal(`/ipfs/${value}`)
```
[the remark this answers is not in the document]
vasco-santos: Done! And also updated on  [js-ipfs#1400](https://github.com/ipfs/js-ipfs/pull/1400)

vasco-santos:
@alanshaw your new comments were fixed, and a comment was left for the `timeout`

js/src/name/publish.js
```diff
@@ -0,0 +1,103 @@
...
+
+          ipfs.files.add(fixture.data, { pin: false }, done)
+        })
+      })
+    })
+
+    after((done) => common.teardown(done))
+
+    it('should publish an IPNS record with the default params', (done) => {
+      this.timeout(50 * 1000)
```
alanshaw: Apologies I've just noticed that a bunch of these tests are using `this.timeout` in an arrow function - you'll need to change it to `function (done) {/*...*/}` for it to be referring to the correct `this` (otherwise you're referring to `this` from the `describe` block)

js/src/name/publish.js
```diff
@@ -0,0 +1,103 @@
...
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+        expect(res.name).to.equal(nodeId)
+        expect(res.value).to.equal(`/ipfs/${value}`)
+
+        done()
+      })
+    })
+
+    it('should recursively resolve to an IPFS hash', (done) => {
```
alanshaw: This test might need to be renamed or rewritten. Was the intention to ensure that publishing `/ipfs/rootHash/path/to/file` will actually result in `/ipfs/fileHash` being published as the value? There doesn't seem to be anything recursive going on here.

js/src/name/resolve.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+
+      ipfs.name.publish(value, publishOptions, (err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+
+        // guarantee that the record has an expired validity.
+        setTimeout(function () {
+          ipfs.name.resolve(nodeId, (err, res) => {
+            expect(err).to.exist()
```
alanshaw: Is there an error message we can assert on here (would have to be the same for go-ipfs and js-ipfs)?

js/src/name/resolve.js
```diff
@@ -0,0 +1,136 @@
...
+        }, 1)
+      })
+    })
+
+    it('should recursively resolve to an IPFS hash', (done) => {
+      this.timeout(100 * 1000)
+
+      const value = fixture.cid
+      const publishOptions = {
+        resolve: true,
```
alanshaw: Shouldn't this be `false` for this test?

js/src/name/publish.js
vasco-santos: **Recursive:** Resolve until the result is not an IPNS name. Default: “false”. Required: no. (according to the docs at https://ipfs.io/docs/api/#apiv0namepublish)

This test aims to do the following:

1) Publish an `ipfs path` with the ID of the peer as the key
Result: `/ipns/{PeerID}` => `/ipfs/{CID}`

2) Publish the previously `ipns path` with the ID of a new key created
Result: `/ipns/{KeyID}` => `/ipns/{PeerID}`

3) Resolve the last recorder IPNS entry recursively:
Result: `/ipfs/{CID}`

In this test, the resolve happens recursively until the stop condition (obtaining an `ipfs path` or reaching a maximum number of tries constant). This way, the first resolve of `/ipns/{KeyID}` will result in `/ipns/{PeerID}`, which is not an `ipfs path`. Accordingly, the resolve continues, and resolving `/ipns/{PeerID}` results in `/ipfs/{CID}`. As we got an `ipfs path`, this is returned.

Maybe the test description is not the most understandable, but for the people who know what the `recursive` parameter aims to, will understand the logic. If you have any suggestion for renaming it, I will happily change it 🙂

js/src/name/resolve.js
vasco-santos: yeah, I am using the same message as `go-ipfs` and I will add an assertion for it.

js/src/name/resolve.js
vasco-santos: It will be the same having it enabled or disabled. In fact, `resolve` is true by default. Accordingly, when a name publish is performed, it is verified that the `ipfs path` / `ipns path` that we want to publish already was added. Otherwise, we may create a `ipns path` for a unavailable file. 

However, as I have other tests with `resolve: true` I will switch it for `false` here, which will result in decreasing the time of the test.

js/src/name/publish.js
vasco-santos: Wrong test name in fact, this was for `name/resolve`

PR #330, opened 2018-07-18
test(files): add test to ensure meaningful error is thrown in `cat()`

0x-r4bbit:
To verify changes discussed in https://github.com/ipfs/js-ipfs-api/issues/799#issuecomment-402086401, this commit
adds a test that checks `files.cat()` throws a meaningful error when a key or ipfs path with
unsupported encoding is given.

License: MIT
Signed-off-by: Pascal Precht pascal.precht@gmail.com

### DIFF
--- a/js/src/files/cat.js
+++ b/js/src/files/cat.js
@@ -130,6 +130,16 @@ module.exports = (createCommon, options) => {
         })
     })
 
+    it('should error on key/path with invalid encoding', () => {
+      const invalidCid = '/ipfs/QmUzdHisA7ouM8DM5rVAKursVK7t7tKeqDYGvmaCdJw5jEeesdlfjsldkfjsldkfjsdlkfjskldfj'
+
+      return ipfs.files.cat(invalidCid)
+        .catch((err) => {
+          expect(err).to.exist()
+          expect(err.toString()).to.contain('Error: selected encoding not supported')
+        })
+    })
+
     it('should error on unknown path (promised)', () => {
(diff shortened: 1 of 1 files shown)


alanshaw:
Hey @PascalPrecht did you run this against js-ipfs as well as js-ipfs-api to check it passes for go-ipfs as well as js-ipfs both node types?

PascalPrecht:
Ha good point @alanshaw !

I've only tested this against js-ipfs-api (where it passes). Running this against `js-ipfs` gives me:

```
  1) interface-ipfs-core tests
       .files.cat
         should error on key/path with invalid encoding:
     AssertionError: expected 'Error: Non-base58 character' to include 'Error: selected encoding not supported'
      at ipfs.files.cat.catch (/Users/pascalprecht/projects/ipfs/interface-ipfs-core/js/src/files/cat.js:139:37)
```

So `js-ipfs` seems to behave differently to `go-ipfs` here as well. At least they throw different errors (while both seem legit). What should do you think should be done? Shall I change the test to just expecting *some* error?

I think though, it'd be better to specify *what* error we're expecting...

PascalPrecht:
@VictorBjelkholm Hi and thank you for getting back to me here!

Sure, I can add that.

Do you also have some thoughts on 

>  What should do you think should be done? Shall I change the test to just expecting some error?

By any chance?

daviddias:
@PascalPrecht can you rebase master onto this branch?

PR #331, opened 2018-07-18
Links to examples are broken

vutsalsinghal:
I've added proper links to examples

### DIFF
--- a/SPEC/FILES.md
+++ b/SPEC/FILES.md
@@ -75,7 +75,7 @@ ipfs.ls(validCID, function (err, files) {
 })
 ```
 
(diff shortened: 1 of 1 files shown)


alanshaw:
Thanks for your PR @vutsalsinghal!

Looks like a lot of these links got broken during the big refactor recently.

Instead of repeating the URL multiple times I think the best fix is to change one line at the bottom of the file:

```md
# Currently:
[examples]: https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/files.js
# Change to:
[examples]: https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/files
```

Leave `[examples][]` as it is.

It would be awesome if you could send a PR to fix the link in the other spec documents as well 🚀 💯

vutsalsinghal:
sure!

PR #332, opened 2018-07-19
test: add tests and docs for ipfs.resolve

alanshaw:
"should resolve IPNS link recursively" takes > 1min on my laptop. I'm open to suggestions on how to speed this up...

refs https://github.com/ipfs/js-ipfs-api/pull/794 and https://github.com/ipfs/js-ipfs/pull/1455

### DIFF
--- a/js/src/miscellaneous/resolve.js
+++ b/js/src/miscellaneous/resolve.js
@@ -35,7 +35,7 @@ module.exports = (createCommon, options) => {
     it('should resolve an IPFS hash', (done) => {
       const content = loadFixture('js/test/fixtures/testfile.txt', 'interface-ipfs-core')
 
-      ipfs.add(content, (err, res) => {
+      ipfs.files.add(content, (err, res) => {
         expect(err).to.not.exist()
         expect(isIpfs.cid(res[0].hash)).to.be.true()
 
@@ -52,7 +52,7 @@ module.exports = (createCommon, options) => {
       const path = '/path/to/testfile.txt'
       const content = loadFixture('js/test/fixtures/testfile.txt', 'interface-ipfs-core')
 
-      ipfs.add([{ path, content }], { wrapWithDirectory: true }, (err, res) => {
+      ipfs.files.add([{ path, content }], { wrapWithDirectory: true }, (err, res) => {
         expect(err).to.not.exist()
 
         const rootHash = res.find(r => r.path === '').hash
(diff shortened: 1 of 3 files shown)


js/src/miscellaneous/resolve.js
```diff
@@ -0,0 +1,112 @@
...
+        ipfs.resolve(path, (err, path) => {
+          expect(err).to.exist()
+          expect(err.message).to.equal('found non-link at given path')
+          done()
+        })
+      })
+    })
+
+    // Test resolve turns /ipns/domain.com into /ipfs/QmHash
+    it('should resolve an IPNS DNS link', function (done) {
```
VictorBjelkholm: This case is a bit worrisome. First we have the reliance on a remote service, secondly we're not actually checking if the output is correct.

My first thought is that in general for the ipns tests, we should have a mini-dns service running together with the tests, that the local daemon can resolve IPNS to. So instead of testing `/ipns/ipfs.io` we should test `/ipns/local-dns-service.home` which would be resolved only locally and we can then assert correct output.

alanshaw: I would like that too. I have no idea how to do that or if it's even possible but I'll look into it. Do you have any pointers?

alanshaw:
@VictorBjelkholm I've added skips to the DNS related tests until we can find a good way to not depend on external DNS records.

I had an idea that for these tests the TXT record to lookup would be at the domain `[cid].ipfs-test.local` and we'd use an MDNS server to respond to the query. The server could look at the CID in the subdomain and would reply with `dnslink=/ipfs/[cid]`.

I tried using https://www.npmjs.com/package/mdns-server for this but it wasn't able to bind to the MDNS ports. I'm guessing IPFS is already bound to them 😕?

Maybe we can setup Jenkins to run a DNS server with some known records?

VictorBjelkholm:
> I've added skips to the DNS related tests until we can find a good way to not depend on external DNS records.

Let's run them for now, if they become a problem we can skip them. Not relying on external DNS is a nice-to-have-feature and not required. We also already have bunch of other tests that are having a dependency on external DNS.

> Maybe we can setup Jenkins to run a DNS server with some known records?

I rather not, as the tests would run different locally. Would be nicer to have some in-process dns server we can startup/shutdown together with the tests. Then we can also easier test what happens when records are changed etc etc.

Created a issue to track it here: https://github.com/ipfs/testing/issues/58

PR #337, opened 2018-07-20
chore(dht): add API to allow options in `findProvs()`

0x-r4bbit:
As discussed in: https://github.com/ipfs/js-ipfs/issues/1322#issuecomment-385336102

License: MIT
Signed-off-by: Pascal Precht <pascal.precht@gmail.com>

### DIFF
--- a/SPEC/DHT.md
+++ b/SPEC/DHT.md
@@ -39,9 +39,12 @@ A great source of [examples][] can be found in the tests for this API.
 
 ##### `Go` **WIP**
 
-##### `JavaScript` - ipfs.dht.findprovs(hash, [callback])
+##### `JavaScript` - ipfs.dht.findprovs(hash, [options], callback])
 
 Where `hash` is a multihash.
+
+`options` an optional object with the following properties
+  - `timeout` - a maximum timeout in milliseconds
 
 `callback` must follow `function (err, peerInfos) {}` signature, where `err` is an error if the operation was not successful. `peerInfos` is an array of objects of type [PeerInfo](https://github.com/libp2p/js-peer-info)
 
@@ -51,6 +54,8 @@ If no `callback` is passed, a promise is returned.
 
 ```JavaScript
 ipfs.dht.findprovs(multihash, function (err, peerInfos) {})
+
+ipfs.dht.findprovs(multihash, { timeout: 4000 }, function (err, peerInfos) {})
 ```
 
 A great source of [examples][] can be found in the tests for this API.
--- a/js/src/dht/findprovs.js
+++ b/js/src/dht/findprovs.js
@@ -67,7 +67,7 @@ module.exports = (createCommon, options) => {
       ], done)
     })
 
-    it('should take options to override timout config', function (done) {
+    it('should take options to override timeout config', function (done) {
       const options = {
         timeout: 1
       }
--- a/js/src/dht/findprovs.js
+++ b/js/src/dht/findprovs.js
@@ -1,11 +1,23 @@
 /* eslint-env mocha */
 'use strict'
 
+const multihashing = require('multihashing-async')
+const Crypto = require('crypto')
 const waterfall = require('async/waterfall')
 const CID = require('cids')
 const { spawnNodesWithId } = require('../utils/spawn')
 const { getDescribe, getIt, expect } = require('../utils/mocha')
 const { connect } = require('../utils/swarm')
+
+function fakeCid (cb) {
+  const bytes = Crypto.randomBytes(Math.round(Math.random() * 1000))
+  multihashing(bytes, 'sha2-256', (err, mh) => {
+    if (err) {
+      cb(err)
+    }
+    cb(null, new CID(0, 'dag-pb', mh))
+  })
+}
 
 module.exports = (createCommon, options) => {
   const describe = getDescribe(options)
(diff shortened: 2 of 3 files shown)


SPEC/DHT.md
```diff
@@ -51,6 +54,8 @@ If no `callback` is passed, a promise is returned.
 
 ```JavaScript
 ipfs.dht.findprovs(multihash, function (err, peerInfos) {})
+
+ipfs.dht.findprovs(multihash, { timeout: 4000 }, function (err, peerInfos) {})
```
PascalPrecht: I'd like to add a test for this, but I'm not sure how to test this timeout.

Appreciate any points, in case anybody wants this as well :)

alanshaw: I'd set the timeout really low and give it a hash that no-one will have, the callback should be called with an error before mocha times out the test.

SPEC/DHT.md
```diff
@@ -39,10 +39,13 @@ A great source of [examples][] can be found in the tests for this API.
 
 ##### `Go` **WIP**
 
-##### `JavaScript` - ipfs.dht.findprovs(hash, [callback])
+##### `JavaScript` - ipfs.dht.findprovs(hash, [options, callback])
```
alanshaw: Nitpick - could this please be changed to `ipfs.dht.findprovs(hash, [options], [callback])`? The other way makes it look like an array which I think is a bit confusing!

PascalPrecht: Of course! I honestly wasn't sure what the right syntax expression would be here. Will update accordingly

SPEC/DHT.md
PascalPrecht: Okay, I'm looking into this now and - sorry if this is a dumb question - but I'm wondering how I should know what hash would be a hash that no one will have? I mean I can't really know this upfront right?

Or is it enough to know, that no one of the connected peers will have it in the testing scenario, which in this case would be `nodeB`, I guess? Considering this already existing testing setup:

```
common.setup((err, factory) => {
        expect(err).to.not.exist()

        spawnNodesWithId(2, factory, (err, nodes) => {
          expect(err).to.not.exist()

          nodeA = nodes[0]
          nodeB = nodes[1]

          connect(nodeB, nodeA.peerId.addresses[0], done)
        })
})
```

What I've tried now, was to create an empty dagNode object within `nodeB` to get hold of its CID, but I don't actually announce the content in the network with `provide()`:

```
it('should take options to override timout config', function (done) {
  const options = {
    timeout: 1
  }

  waterfall([
    (cb) => nodeB.object.new(cb),
    (dagNode, cb) => {
      // Notice I'm not announcing the hash in the network using `provide()` but simply return 
      // CID
      const cidV0 = new CID(dagNode.toJSON().multihash)
      cb(cidV0)
    },
    (cidV0, cb) => nodeA.dht.findprovs(cidV0, options, (err) => {
      // I'd expect this to be truthy as nodeA shouldn't find a provider for the given hash
      expect(err).to.exist()
      cb(err)
    }),
  ], done) // this fails due to non-error being passed to done
})

```
So this test fails because `done()` doesn't get called with an error object, which means that `findProvs` doesn't throw and error here. 

Can nodes find providers, even though other nodes didn't announce their content to the network?

alanshaw: I think that by adding content via `object.new` that you become a provider for that CID automatically.

We have a helper for the webui tests that creates a CID from random bytes without adding content to IPFS. Perhaps something similar here might work?

https://github.com/ipfs-shipyard/ipfs-webui/blob/revamp/test/helpers/cid.js

PascalPrecht: Yes, that could work. I'll give it a spin

js/src/dht/findprovs.js
```diff
@@ -1,12 +1,25 @@
 /* eslint-env mocha */
 'use strict'
 
+const multihashing = require('multihashing-async')
+const promisify = require('util').promisify
```
alanshaw: Unused?

js/src/dht/findprovs.js
```diff
@@ -1,12 +1,25 @@
...
 const { getDescribe, getIt, expect } = require('../utils/mocha')
 const { connect } = require('../utils/swarm')
 
+function fakeCid (cb) {
+  const bytes = Crypto.randomBytes(Math.round(Math.random() * 1000))
+  multihashing(bytes, 'sha2-256', (err, mh) => {
+    if (err) {
+      cb(err)
+    }
+    cb(new CID(0, 'dag-pb', mh))
```
alanshaw: `cb(null, new CID(0, 'dag-pb', mh))`

js/src/dht/findprovs.js
```diff
@@ -54,5 +67,18 @@ module.exports = (createCommon, options) => {
...
+        timeout: 1
+      }
+      waterfall([
+        (cb) => fakeCid(cb),
+        (cidV0, cb) => nodeA.dht.findprovs(cidV0, options, (err) => {
+          expect(err).to.exist()
+          cb(null)
+        }),
+      ], done)
+    });
```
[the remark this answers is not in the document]
alanshaw: The issue is that `fakeCid` returns a CID in the error position of the callback

alanshaw (changes requested):

js/src/dht/findprovs.js
PascalPrecht: Ah! Thanks!! 

And also... D'oh! 🤦‍♂️

js/src/dht/findprovs.js
PascalPrecht: Yeap, that's a left over

PascalPrecht:
@alanshaw thanks again for taking a look.

This is now updated and passes tests on js-ipfs.

PR #342, opened 2018-08-02
test: make files.ls results ordering consistent

achingbrain:
Previously we'd get files in different orders between `files.ls()` and `files.ls({long: true})` due to a bug in `go-ipfs` (see ipfs/go-ipfs#5181)
    
Since this has been resolved we can now expect consistent ordering.


achingbrain:
This will break the `js-ipfs` tests until ipfs/js-ipfs-mfs#6 is merged & released to restore compatibility on the js side.

PR #343, opened 2018-08-03
docs: change config.get() to object

hacdias:
Ref.: https://github.com/ipfs/js-ipfs-api/issues/822

### DIFF
--- a/SPEC/CONFIG.md
+++ b/SPEC/CONFIG.md
@@ -14,7 +14,7 @@
 
 `key` is the key of the value that should be fetched from the config file. If no key is passed, then the whole config should be returned. `key` should be of type String.
 
-`callback` must follow `function (err, config) {}` signature, where `err` is an error if the operation was not successful and `config` is a JSON object containing the configuration of the IPFS node.
+`callback` must follow `function (err, config) {}` signature, where `err` is an error if the operation was not successful and `config` is an object containing the configuration of the IPFS node.
 
 If no callback is passed, a [promise][] is returned
 


hacdias:
@diasdavid @alanshaw what do you think? 😄

SPEC/CONFIG.md
```diff
@@ -14,7 +14,7 @@
 
 `key` is the key of the value that should be fetched from the config file. If no key is passed, then the whole config should be returned. `key` should be of type String.
 
-`callback` must follow `function (err, config) {}` signature, where `err` is an error if the operation was not successful and `config` is a JSON object containing the configuration of the IPFS node.
+`callback` must follow `function (err, config) {}` signature, where `err` is an error if the operation was not successful and `config` is an object containing the configuration of the IPFS node.
```
diasdavid: Are you saying that the config is not pure JSON?

hacdias: The config is a JavaScript object, which can be stringified to JSON.

Please see https://github.com/ipfs/js-ipfs-api/issues/822 and https://github.com/ipfs/js-ipfs-api/pull/825 😄

diasdavid: It should be an object everywhere. What we mean by it being a JSON object is that it should include __no buffers__ on it.

hacdias: @diasdavid Oh, makes sense then. I thought "JSON object" == "Buffer with JSON string". So this PR can be closed, but https://github.com/ipfs/js-ipfs-api/pull/825 should be merged then.

PR #349, opened 2018-08-09
docs: document --parents flag for files.write command

achingbrain:
Also adds a test.

Already supported in `js-ipfs`, `go-ipfs` requires ipfs/go-ipfs#5359 to be merged and released

### DIFF
--- a/js/src/files/write.js
+++ b/js/src/files/write.js
@@ -40,21 +40,31 @@ module.exports = (createCommon, options) => {
       })
     })
 
-    it('should write to non existent file with create flag, expect no error', function (done) {
-      const testDir = `/test-${hat()}`
+    it('should write to non existent file with create flag', function (done) {
+      const testPath = `/test-${hat()}`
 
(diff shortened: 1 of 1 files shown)


js/src/files/write.js
```diff
@@ -48,5 +48,14 @@ module.exports = (createCommon, options) => {
         done()
       })
     })
+
+    it('should write to deeply nested non existent file with create and parents flags, expect no error', function (done) {
```
VictorBjelkholm: This test doesn't actually test if a parent directory is being created, only that the function doesn't error

achingbrain: That's true.  Neither do the other tests present, nor do the related `go-ipfs` sharness tests.

achingbrain: 😢

VictorBjelkholm: Looks better now though! Thanks for fixing it + the other test

PR #357, opened 2018-08-28
Added a list of all possible hash algorithms to `hash` option as well as default hash function

aleybovich:


alanshaw:
Thanks @aleybovich

PR #360, opened 2018-08-31
WIP: fix: ipns tests using go daemon

vasco-santos:
In the context of having the `ipns` tests running on [js-ipfs-api#841](https://github.com/ipfs/js-ipfs-api/pull/841)

### DIFF
--- a/js/src/name/resolve.js
+++ b/js/src/name/resolve.js
@@ -41,7 +41,7 @@ module.exports = (createCommon, options) => {
     after((done) => common.teardown(done))
 
     it('should resolve a record with the default params after a publish', function (done) {
-      this.timeout(50 * 1000)
+      this.timeout(120 * 1000)
 
       const value = fixture.cid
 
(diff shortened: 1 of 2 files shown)


js/src/name/resolve.js
```diff
@@ -125,7 +123,7 @@ module.exports = (createCommon, options) => {
             ipfs.name.resolve(keyId, resolveOptions, (err, res) => {
               expect(err).to.not.exist()
               expect(res).to.exist()
-              expect(res.path).to.equal(`/ipfs/${value}`)
+              expect(res.path || res).to.equal(`/ipfs/${value}`) // TODO
```
VictorBjelkholm: This seems wrong, tests should test one thing and here is a conditional. Also the `TODO` seems to be missing a comment about what/why

js/src/name/resolve.js
```diff
@@ -83,12 +81,12 @@ module.exports = (createCommon, options) => {
 
             done()
           })
-        }, 1)
+        }, 10)
```
VictorBjelkholm: Rather than having this is a timeout, we should have something similar to this: https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/pubsub/utils.js#L5-L31

Where we execute the expectations once every 100 milliseconds until either it passes or we've tried 10 times. This will make the test less likely to fail on lower performance machines.

js/src/name/resolve.js
vasco-santos: yes! That will be changed 🙂

But for explaining the purpose, I am experiencing different responses when the daemon is online and offline. One has the path property, while the other does not have. I will keep this in `WIP` before solving this

PR #361, opened 2018-08-31
test: ipns over pubsub

vasco-santos:
PR implemented taking into consideration the most recent documentation of the `go-ipfs` API on:

[http://ipfs.io/ipfs/QmPKzWaeHumLuiKLb2Ak67s5kxZ1p6YbnjVKYYfF9hTjzn](http://ipfs.io/ipfs/QmPKzWaeHumLuiKLb2Ak67s5kxZ1p6YbnjVKYYfF9hTjzn)

This tests aim to be integrated on [s-ipfs-api#846](https://github.com/ipfs/js-ipfs-api/pull/846) and in a future PR to `js-ipfs` containing the `IPNS over Pubsub` implementation.


SPEC/NAME.md
```diff
@@ -76,6 +79,8 @@ This way, you can republish a new version of your website under the same address
 
 `callback` must follow `function (err, name) {}` signature, where `err` is an error if the operation was not successful. `name` is a string that contains the IPFS hash.
 
+`callback` must follow `function (err, result) {}` signature, where `err` is an error if the operation was not successful. `result` is an object that contains the resulting path, such as:
```
diasdavid: image

Why are there two callback definitions?

SPEC/NAME.md
```diff
@@ -2,6 +2,9 @@
 
 * [name.publish](#namepublish)
 * [name.resolve](#nameresolve)
+* [name.pubsub.cancel](#namepubsubcancel)
+* [name.pubsub.state](#namepubsubstate)
+* [name.pubsub.subs](#namepubsubsubs)
```
alanshaw: Could we re-order these so they're alphabetical with the others? I know this isn't the case for other spec docs but it's something I'd like to start doing. It just makes it easier to find docs for a particular method if you already know it's name.

js/src/name-pubsub/cancel.js
```diff
@@ -0,0 +1,80 @@
...
+    })
+
+    it('should cancel a subscription correctly returning true', function (done) {
+      this.timeout(140 * 1000)
+      const value = fixture.cid
+
+      ipfs.name.publish(value, { resolve: false }, (err, res) => {
+        expect(err).to.not.exist()
+
+        ipfs.name.resolve(nodeId, (err) => {
```
alanshaw: Just so I understand this correctly, when you resolve, you get automatically subscribed to updates? Is there a way to not subscribe?

js/src/name-pubsub/cancel.js
```diff
@@ -0,0 +1,80 @@
...
+          expect(err).to.not.exist()
+
+          ipfs.name.pubsub.cancel(nodeId, (err, res) => {
+            expect(err).to.not.exist()
+            expect(res).to.exist()
+            expect(res).to.have.property('canceled')
+            expect(res.canceled).to.eql(true)
```
alanshaw: Should we also check this value is not longer in our subs list?

js/src/name-pubsub/cancel.js
```diff
@@ -0,0 +1,80 @@
...
+
+            done()
+          })
```
alanshaw: Suggestion: async waterfall or something?

js/src/name-pubsub/cancel.js
```diff
@@ -0,0 +1,80 @@
...
+'use strict'
+
+const loadFixture = require('aegir/fixtures')
+
+const { spawnNodeWithId } = require('../utils/spawn')
+const { getDescribe, getIt, expect } = require('../utils/mocha')
+
+const fixture = Object.freeze({
+  data: loadFixture('js/test/fixtures/testfile.txt', 'interface-ipfs-core'),
+  cid: 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
```
alanshaw: IMHO we should just use the CID in the result from add and not hard code it. We're not testing "add" functionality here and if the CID changes in the future (because of changes in default dag layout or cid version for instance) there's fewer things to update in the tests. If you do change this then please also change in the other test files :)

js/src/name-pubsub/subs.js
```diff
@@ -0,0 +1,80 @@
...
+    after((done) => common.teardown(done))
+
+    it('should get a null result of subscriptions before any resolve', function (done) {
+      this.timeout(60 * 1000)
+
+      ipfs.name.pubsub.subs((err, res) => {
+        expect(err).to.not.exist()
+        expect(res).to.exist()
+        expect(res).to.have.property('strings')
+        expect(res.strings).to.eql(null)
```
alanshaw: It should always be an array.

js/src/name-pubsub/subs.js
```diff
@@ -0,0 +1,80 @@
...
+        expect(err).to.not.exist()
+
+        ipfs.name.resolve(nodeId, (err) => {
+          expect(err).to.not.exist()
+
+          ipfs.name.pubsub.subs((err, res) => {
+            expect(err).to.not.exist()
+            expect(res).to.exist()
+            expect(res).to.have.property('strings')
+            expect(res.strings).to.be.an('array').that.does.include(`/ipns/${nodeId}`)
```
alanshaw: Check that it was empty initially?

js/src/name-pubsub/subs.js
```diff
@@ -0,0 +1,80 @@
...
+
+            done()
+          })
```
alanshaw: Suggestion: use async

SPEC/NAME.md
```diff
@@ -84,8 +87,93 @@ If no `callback` is passed, a promise is returned.
...
+
+##### `Go` **WIP**
+
+##### `JavaScript` - ipfs.name.pubsub.subs([callback])
+
+`callback` must follow `function (err, result) {}` signature, where `err` is an error if the operation was not successful. `result` is an object that contains the result of the operation, such as:
+
+```JavaScript
+{
+  strings: ['/ipns/QmQrX8hka2BtNHa8N8arAq16TCVx5qHcb46c5yPewRycLm']
```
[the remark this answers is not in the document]
vasco-santos: Yes! I got confused here. Both js-ipfs-api and js-ipfs should return only the array in the end!

js/src/name-pubsub/cancel.js
vasco-santos: When you try to `resolve` an ipns path and the experimental flag `--enable-namesys-pubsub` is being used, it will get automatically subscribed. This way, when an ipns record is updated, we will have the most recent version of it stored locally.

For now, there is no way to not subscribe it (if you use the experimental flag), but since it is experimental, we may consider adding a flag for it, in case we enable it by default in the future.

js/src/name-pubsub/cancel.js
vasco-santos: Will change in both

SPEC/NAME.md
vasco-santos: done!

PR #367, opened 2018-09-17
fix: add test for block rm

vmx:
Sadly those tests can't be enabled yet as js-ipfs-api doesn't implement `block/rm` yet (see https://github.com/ipfs/js-ipfs-api/issues/792).

But I thought I make a pull request with the test though, so that work doesn't get wasted.

### DIFF
--- a/js/src/block/rm.js
+++ b/js/src/block/rm.js
@@ -1,0 +1,54 @@
+/* eslint-env mocha */
+'use strict'
+
+const CID = require('cids')
+const auto = require('async/auto')
+const { getDescribe, getIt, expect } = require('../utils/mocha')
+
+module.exports = (createCommon, options) => {
+  const describe = getDescribe(options)
+  const it = getIt(options)
+  const common = createCommon()
+
+  describe('.block.rm', function () {
+    const data = Buffer.from('blorb')
+    let ipfs, hash
+
+    before(function (done) {
+      // CI takes longer to instantiate the daemon, so we need to increase the
+      // timeout for the before step
+      this.timeout(60 * 1000)
+
+      auto({
+        factory: (cb) => common.setup(cb),
+        ipfs: ['factory', (res, cb) => res.factory.spawnNode(cb)],
+        block: ['ipfs', (res, cb) => res.ipfs.block.put(data, cb)]
+      }, (err, res) => {
+        if (err) return done(err)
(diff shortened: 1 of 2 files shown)


js/src/block/rm.js
```diff
@@ -0,0 +1,54 @@
...
+      ipfs.block.rm(cid, (err) => {
+        expect(err).to.not.exist()
+        done()
+      })
+    })
+
+    it('should error on removing non-existent block', (done) => {
+      const cid = new CID('QmYi5NFboBxXvdoRDSa7LaLcQvCukULCaDbZKXUXz4umPa')
+      ipfs.block.rm(cid, (err, block) => {
+        expect(err).to.exist()
```
VictorBjelkholm: We should make sure the error message contains whatever we expect it to have inside of it.

js/src/block/rm.js
```diff
@@ -0,0 +1,54 @@
...
+      })
+    })
+
+    after((done) => common.teardown(done))
+
+    it('should remove by CID object', (done) => {
+      const cid = new CID(hash)
+      ipfs.block.rm(cid, (err) => {
+        expect(err).to.not.exist()
+        done()
```
VictorBjelkholm: We need to make sure the block was removed, but I'm not sure if we have any public API for checking if a block already exists locally?

alanshaw: Can we use https://docs.ipfs.io/reference/api/http/#api-v0-refs-local for this check?

js/src/block/rm.js
alanshaw: +1

vmx:
@alanshaw The problem is that this change has a looooong tail. IIRC things are not implemented as expected/correctly. But I'll have a another look at the test, it's worth adding it even if we need to skip it atm.

alanshaw:
I'm happy to add skips where appropriate as long as the tests are passing in at least one of js-ipfs or js-ipfs-api 🚀

diasdavid:
block.rm was deleted from the API as it messes with Pinning for go-ipfs. Read https://github.com/ipfs/js-ipfs-api/issues/792

achingbrain:
Slightly more comprehensive tests were added in #487 and the implementation will ship with `ipfs@0.39.0`

PR #374, opened 2018-10-09
chore: remove ipld formats re-export

vmx:
Prior to this change the `ipld-dag-cbor` and `ipld-dag-pb` modules
are re-exported so that can be accessed within the Browser bundle.
Those modules normally don't need to be used directly, they are
kind of implementation details of IPLD. Hence remove them.

### DIFF
--- a/js/src/types.js
+++ b/js/src/types.js
@@ -3,8 +3,6 @@
 
 const PeerId = require('peer-id')
 const PeerInfo = require('peer-info')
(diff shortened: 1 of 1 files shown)


vmx:
Those three commits should be merged together:

 - https://github.com/ipfs/js-ipfs/pull/1626
 - https://github.com/ipfs/js-ipfs-api/pull/872
 - https://github.com/ipfs/interface-ipfs-core/pull/374

PR #376, opened 2018-10-23
test: additional add* validation tests

### LINKED ISSUE #292: files.add*Stream coverage is poor
Pulling out tests per command for https://github.com/ipfs/interface-ipfs-core/pull/290 shows that test coverage for `addReadableStream` and `addPullStream` is poor (almost non existent).

Considering `files.add` is basically most developer's first port of call when starting with IPFS these commands should be rock solid and tested as such.

alanshaw:
refs #292

### DIFF
--- a/js/src/files-regular/add.js
+++ b/js/src/files-regular/add.js
@@ -120,6 +120,26 @@ module.exports = (createCommon, options) => {
       })
     })
 
+    it('should not be able to add a string', (done) => {
+      const data = `TEST${Date.now()}`
+
+      ipfs.files.add(data, (err) => {
+        expect(err).to.exist()
+        expect(err.message).to.contain('invalid input')
+        done()
+      })
+    })
+
+    it('should not be able to add a non-Buffer TypedArray', (done) => {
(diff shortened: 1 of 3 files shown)


js/src/files/add.js
```diff
@@ -120,6 +120,26 @@ module.exports = (createCommon, options) => {
...
+      const data = `TEST${Date.now()}`
+
+      ipfs.files.add(data, (err) => {
+        expect(err).to.exist()
+        expect(err.message).to.contain('invalid input')
+        done()
+      })
+    })
+
+    it('should not be able to add a non-Buffer TypedArray', (done) => {
```
VictorBjelkholm: Should this test include more of the different typed arrays?

image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays#Typed_array_views

alanshaw: I think that since these all inherit from `TypedArray` then testing with just one is good enough, the point here is to ensure the API is feeding back with an appropriate message and not passing the input further down the stack where it'll cause an error that doesn't communicate the real problem.

Issue #118, opened 2017-02-16
Consider enabling `add(path)` to the files API


daviddias:
@jbenet brought up that a lot of users will expect `ipfs.add(path)` to just work as if they were in the terminal. 

In the past, we went through several iterations and discussions and eventually ended up agreeing that it would be beneficial to have utilities like [addFromFs and addFromUrl](https://github.com/ipfs/js-ipfs-api#add-files-or-entire-directories-from-the-filesystem-to-ipfs) as actions that are explicit, this keeps the promise intact that code can be run in the browser and in Node.js, as long as the dev is using the ipfs-core API. 

We can add this feature, we will just need to be verbose about that caveat.

dignifiedquire:
👎 on this, it makes the api more complicated, and having an api call that has entirely different semantics depending on if it is run in node.js or the browser seems like it will only cause more confusion.

jbenet:
This is a basic expectation of ipfs nodes and tooling.

- I would expect `ipfs.add(path)` to work, everywhere, and only fail when the path is inaccessible.
- in the browser, the limitation is the access to the files, similar to how in the filesystem permissions may hide parts of the fs. `ipfs.add("~/foo/bar")` is not possible in the browser because the files are not accessible.
- you could choose to make `ipfs.add(url)` work in the browser, as i would expect to work. (http is sort of the fs of the web.)

Bottom line, these should be trivially easy to use:

```js
// direct analogues with ipfs add in the cli, in node.
ipfs.add("~/path/to/my-file")
ipfs.add("~/path/to/my-dir", {recursive: true})
ipfs.add("my-file")
ipfs.add("my-dir", {recursive: true})
ipfs.add(stream) // makes a file

// direct analogues with ipfs add in the cli, but on the browser
ipfs.add("https://foo.com/my-page")
// arguably, people would expcet this to work like wget mirroring:
// ipfs.add("https://foo.com/my-page", {recursive: true})
// but i'm fine not supporting it.
```

Conversely, i would expect cat to work equally easily:

```js
ipfs.cat(ipfsPath) # returns a stream
```

### The emphasis here is in:

- consistency with ipfs tooling + expectations everywhere else
- trivially simple command for trivially simple use case
- maintaining backwards compatibility. [don't break interfaces.](https://felipec.wordpress.com/2013/10/07/the-linux-way/)

jbenet:
[...]

PR #378, opened 2018-10-25
feat: move regular files api to top level, add addFromFs and addFromUrl

diasdavid:
Related to https://github.com/ipfs/interface-ipfs-core/issues/118


SPEC/FILES.md
```diff
@@ -1,141 +1,41 @@
...
+- [addFromUrl](#addFromFs)
+- [cat](#cat)
+- [catReadableStream](#filescatreadablestream)
+- [catPullStream](#filescatpullstream)
+- [get](#filesget)
+- [getReadableStream](#filesgetreadablestream)
+- [getPullStream](#filesgetpullstream)
 - [ls](#ls)
 - [lsReadableStream](#lsreadablestream)
 - [lsPullStream](#lspullstream)
```
diasdavid: Moving all the non MFS commands to the top level. This preps the users for the future Files API as they will get used to call the add, cat, get and ls from there, but when the MFS parts get coalesced with the non MFS parts, they will just get the lovely surprise that there will be a "drive" looking thing that contains all the files they added.

SPEC/FILES.md
```diff
@@ -1,141 +1,41 @@
...
 
-> The files API enables users to use the File System abstraction of IPFS.
-
+> The files API enables users to use the File System abstraction of IPFS. There are two Files API, one at the top level, the original `add`, `cat`, `get` and `ls`, and another behind the [`files`, also known as MFS](https://github.com/ipfs/specs/issues/98). [We are currently going through a revamping process of this APIs to make them more user friendly](https://github.com/ipfs/interface-ipfs-core/issues/284).
+
+- [add](#add)
+- [addReadableStream](#addreadablestream)
+- [addPullStream](#addpullstream)
+- [addFromFs](#addFromFs)
+- [addFromUrl](#addFromFs)
```
diasdavid: Added the addFromFs (no one should have to do this: https://github.com/ipfs/js-ipfs/blob/master/src/cli/commands/files/add.js#L18-L102) and addFromUrl. 

Once ipfsx is ready, we can squash all these 5 adds into one function.

alanshaw:
Yep, and IPFSX already has a unified `ls`:
https://github.com/alanshaw/ipfsx/pull/1
https://github.com/alanshaw/ipfsx/blob/master/API.md#ls

achingbrain:
I think we should jump straight to the ideal API, rather than make users update their code twice - once to this and then again to ipfsx.

diasdavid:
@achingbrain even if that means 4+ months?

In order for ipfsx to land, we will need:

- (A) Refactor the whole codebase plus test that a stream muxer powered by async generators will be equally performant as the one we have with Streams.
- (B) Agree on the de-facto files API

I believe that we should do these in parallel

alanshaw:
I'm +1 on moving the files API stuff to the top level. IPFS's prime purpose is to deal with files, so these calls should be privileged. They are also the most widely used calls so saving on typing helps developers. WIN WIN.

I think that `addFromURL` and `addFromFs` are convenient right now, but these operations will become really easy if `add` starts to take an async iterator like IPFSX does and so I don't think they'll be needed.

Right now in IPFSX if I want to add these things I'd do something like:

```js
// Add a file
const stream = fs.createReadStream('/path/to/file')
const { cid } = await node.add(stream).first()

// Add a URL
const response = await fetch('http://giffatron.com/cat.gif')
const { cid } = await node.add(response.body.getReader()).first()
```

Ok so, caveat, browser `ReadableStream` isn't a async iterator quite yet but we can shim temporarily.

diasdavid:
@alanshaw what about users that want to add entire directories from their local file system in Node.js (analogous to the CLI ipfs add)? Even with ipfsx, you would still need to do the globbing manually that is happening here https://github.com/ipfs/js-ipfs/blob/master/src/cli/commands/files/add.js#L18-L102 every single time a user is using a embedded node.

hugomrdias:
I agree with @achingbrain we should not do a breaking change just to do another in a few months , I don't believe there sufficient value in it. We can improve docs and make small tweaks to improve experience though. 
I think we should aim for ipfsx I looked into it just now for a second time didn't know it was already so complete. Looks freaking awesome @alanshaw **please** tell me how I can help !! We need this 🎉🎉🎉

alanshaw:
> @alanshaw what about users that want to add entire directories from their local file system in Node.js (analogous to the CLI ipfs add)?

Good point, I didn't consider that.

diasdavid:
@hugomrdias here is a good milestone towards that. Convert https://github.com/libp2p/js-libp2p-mplex to async-generators and run the stream muxer tests on it https://github.com/libp2p/interface-stream-muxer/tree/master/src and compare it with the current implementation in terms of speed and memory usage.

diasdavid:
Can I get a quick vote? I want to make sure to understand if people agree/disagree if people want to ship A and B -- https://github.com/ipfs/interface-ipfs-core/pull/378#issuecomment-432970995 -- in one swipe.

- If you think that we should do A & B from https://github.com/ipfs/interface-ipfs-core/pull/378#issuecomment-432970995 refactor in **parallel**, 🎉 this comment
- If you think that we should do A & B from https://github.com/ipfs/interface-ipfs-core/pull/378#issuecomment-432970995 refactor **simultaneously**, 😄 this comment.

alanshaw:
> I think we should aim for ipfsx I looked into it just now for a second time didn't know it was already so complete. Looks freaking awesome @alanshaw **please** tell me how I can help !! We need this 🎉🎉🎉

❤️ 

There's so much to do. I'm trying to create RFC issues for everything now so that there can be some discussion before an implementation lands. The biggest contribution you can make is using it, suggesting changes, and proposing things you'd like to see.

Also, is there any opposition to moving the ipfsx repo to ipfs shipyard or is it too soon? I think by moving it to shipyard the community might feel it has some longevity and be more open to try it out and it'll also aid discovery.

diasdavid:
From IRC (important for context on the vote https://github.com/ipfs/interface-ipfs-core/pull/378#issuecomment-432976852):

image

SPEC/FILES.md
```diff
@@ -1,141 +1,41 @@
...
+- [cat](#cat)
+- [catReadableStream](#filescatreadablestream)
+- [catPullStream](#filescatpullstream)
+- [get](#filesget)
+- [getReadableStream](#filesgetreadablestream)
+- [getPullStream](#filesgetpullstream)
```
alanshaw: `#filescatreadablestream` => `#catreadablestream`...and others ;)

diasdavid:
Have it all working on https://github.com/ipfs/js-ipfs-api/pull/878. Now onto js-ipfs.

@alanshaw @hugomrdias @achingbrain Please review meanwhile the docs and tests in this PR. Thank you!

PR #382, opened 2018-10-26
chore: delete codeowners

diasdavid:
I don't believe that this file is stating the current reality nor do I know that we effectively use this as part of our contributing guidelines.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,0 @@
-# Please see https://help.github.com/articles/about-codeowners/ for more information
-
-# JavaScript Core Interface Tests
-/js/src/ @VictorBjelkholm @travisperson


alanshaw:
CODEOWNERS is for volunteering yourself to be requested for review when code you added/edited is changed. So, it isn't about documenting who owns the code.

When you change some code what was last modified by someone in the CODEOWNERS file they will **automatically** be requested for a review. There's no pressure to actually do the review, they just get a notification.

The idea is that they will know about the code that is being changed and so should be well qualified to review the change.

This is good because it eleviates some of the burden of review from the lead maintainer and potentially will get the contributor feedback sooner than waiting for the lead maintainer to get round to it.

See https://help.github.com/articles/about-codeowners/ for more.

I would like for JS Core Devs to add themselves to a CODEOWNERS file in other projects too. I often don't know who to request for a review so I just request you.

Do you think this is a good idea? If so I can send a PR to the JS guidelines...

diasdavid:
@alanshaw I think that in theory is good, in practice it will just be another information point that will quickly go out of date.

That said, do PR to the JS Code Guidelines and let the rest of the team comment on it. For now I'll merge this PR as this codeowners file is misleading as neither Victor or Travis are paying attention to this repo to review PRs.

PR #383, opened 2018-10-29
fix: update dht tests

vasco-santos:
In this PR, a currently broken test is skipped and a new test was added to both `dht.get` and `dht.put`.

In the context of [js-ipfs#856](https://github.com/ipfs/js-ipfs/pull/856) and more precisely [js-ipfs#856#issuecomment-435507070](https://github.com/ipfs/js-ipfs/pull/856#issuecomment-435507070).

Needs:

- [x] [ipfs/js-ipfs#856](https://github.com/ipfs/js-ipfs/pull/856)


vasco-santos:
`js-ipfs-api` is ok, but `js-ipfs` will need the [ipfs/js-ipfs#856](https://github.com/ipfs/js-ipfs/pull/856), which will need the next `libp2p` release first. Added a proper label here.

PR #384, opened 2018-10-30
Match new 'no link' error in files.cat

magik6k:
There was a ton of refactoring in go-ipfs which caused a slightly different error to be exposed here


magik6k:
Closing for now as thin may not be needed - https://github.com/ipfs/go-path/pull/12

PR #385, opened 2018-10-30
test: fix resolve test

Stebalien:
1. This is an IPLD path.
2. Resolve really should be resolving up to the last path, not just "failing".

*Neither* js-ipfs *nor* go-ipfs pass this test. However, IMO, a PR is the easiest way to highlight this issue. Thoughts? Objections?

### DIFF
--- a/src/miscellaneous/resolve.js
+++ b/src/miscellaneous/resolve.js
@@ -88,17 +88,17 @@ module.exports = (createCommon, options) => {
       })
     })
 
-    it('should not resolve an IPFS path non-link', (done) => {
+    it('should resolve up to the last node', (done) => {
       const content = { path: { to: { file: hat() } } }
       const options = { format: 'dag-cbor', hashAlg: 'sha2-256' }
 
       ipfs.dag.put(content, options, (err, cid) => {
         expect(err).to.not.exist()
 
-        const path = `/ipfs/${cid.toBaseEncodedString()}/path/to/file`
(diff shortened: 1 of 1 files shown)


js/src/miscellaneous/resolve.js
```diff
@@ -72,17 +72,17 @@ module.exports = (createCommon, options) => {
...
       const options = { format: 'dag-cbor', hashAlg: 'sha2-256' }
 
       ipfs.dag.put(content, options, (err, cid) => {
         expect(err).to.not.exist()
 
-        const path = `/ipfs/${cid.toBaseEncodedString()}/path/to/file`
-        ipfs.resolve(path, (err, path) => {
-          expect(err).to.exist()
-          expect(err.message).to.equal('found non-link at given path')
+        const path = `/ipld/${cid.toBaseEncodedString()}/path/to/file`
```
magik6k: AFAIK js-ipfs doesn't support `/ipld` paths (please tell me I'm wrong). This is probably going to be painful to fix, and blocking go-ipfs release on this is less than ideal.

Stebalien: You're probaly not wrong.

alanshaw:
@Stebalien why is it useful for it to partially resolve a path? Could we implement this as an option `parital=true` to avoid a breaking change?

If we want to merge this, we should have a PR ready for go-ipfs/js-ipfs (or both) as well so that we know the test works!

PR #386, opened 2018-10-31
fix: remove antipattern from ping tests

diasdavid:


diasdavid:
for https://github.com/ipfs/js-ipfs/pull/1685

PR #388, opened 2018-11-08
fix: updates ipld-dag-pb dep to version without .cid or .multihash properties

achingbrain:
Follows on from https://github.com/ipld/js-ipld-dag-pb/pull/99 and updates this module to not rely on DAGNodes having knowledge of their CIDs.


package.json
```diff
@@ -6,6 +6,9 @@
   "main": "js/src/index.js",
   "scripts": {
     "test": "exit 0",
+    "test:node": "exit 0",
+    "test:browser": "exit 0",
+    "test:webworker": "exit 0",
```
achingbrain: These are so Jenkins doesn't fail the build for not having tests.

diasdavid:
I believe that (unfortunately) there are a considerable number of users using this API, perhaps due to a lack of "deprecation warning" 

> As a suggestion, (still a breaking change) but why don't we change the write methods of the object API to return a CID? Then at least users can retrieve a DAGNode using object.get without having to require 'ipld-dag-pb' and calculate their own CID...

It seems to me that this approach + a good deprecation plan is the way to go.

achingbrain:
> I believe that (unfortunately) there are a considerable number of users using this API, perhaps due to a lack of "deprecation warning"

Also because the DAG API isn't implemented over http so you can only use it in JS if you are calling directly into core.

diasdavid:
> Also because the DAG API isn't implemented over http so you can only use it in JS if you are calling directly into core.

There is a way to overcome this - https://github.com/ipfs/js-ipfs-api/pull/755#discussion_r204647385

alanshaw:
> It seems to me that this approach + a good deprecation plan is the way to go.

* 0.34 breaking change to return CIDs
* 0.35 deprecation notices to the release info, blog posts and to the code with something like [`deprecate`](https://www.npmjs.com/package/deprecate)
* 0.X? remove as https://github.com/ipfs/js-ipfs/issues/1670 lands

Sounds good?

achingbrain:
Can we merge this one & I'll put a PR together for returning CIDs from the Object APIs?

achingbrain:
..The reason is I'd like to use a pre-release IPFS on npm-on-ipfs in order to verify that this alleviates the CPU pressure.

PR #389, opened 2018-11-09
fix: update dht responses

vasco-santos:
In the context of the [ipfs/js-ipfs#856](https://github.com/ipfs/js-ipfs/pull/856) and [ipfs/js-ipfs-api/pull/890](https://github.com/ipfs/js-ipfs-api/pull/890)

### DIFF
--- a/js/src/dht/findpeer.js
+++ b/js/src/dht/findpeer.js
@@ -38,10 +38,9 @@ module.exports = (createCommon, options) => {
     after((done) => common.teardown(done))
 
     it('should find other peers', (done) => {
-      nodeA.dht.findpeer(nodeB.peerId.id, (err, peer) => {
+      nodeA.dht.findpeer(nodeB.peerId.id, (err, res) => {
         expect(err).to.not.exist()
-        // TODO upgrade the answer, format is weird
-        expect(peer[0].Responses[0].ID).to.be.equal(nodeB.peerId.id)
+        expect(res.Responses[0].Addrs).to.deep.include(nodeB.peerId.addresses[0])
         done()
       })
     })
--- a/js/src/dht/findpeer.js
+++ b/js/src/dht/findpeer.js
@@ -40,7 +40,8 @@ module.exports = (createCommon, options) => {
     it('should find other peers', (done) => {
       nodeA.dht.findpeer(nodeB.peerId.id, (err, res) => {
         expect(err).to.not.exist()
-        expect(res.Responses[0].Addrs).to.deep.include(nodeB.peerId.addresses[0])
+        expect(res.responses[0].id).to.be.eql(nodeB.peerId.id)
+        expect(res.responses[0].addrs).to.deep.include(nodeB.peerId.addresses[0])
         done()
       })
     })
--- a/js/src/dht/findprovs.js
+++ b/js/src/dht/findprovs.js
@@ -60,7 +60,7 @@ module.exports = (createCommon, options) => {
         },
         (cidV0, cb) => nodeA.dht.findprovs(cidV0, cb),
         (provs, cb) => {
-          expect(provs.Responses.map((p) => p.ID))
+          expect(provs.responses.map((p) => p.id))
             .to.eql([nodeB.peerId.id])
           cb()
         }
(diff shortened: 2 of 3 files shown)


js/src/dht/findpeer.js
```diff
@@ -38,10 +38,9 @@ module.exports = (createCommon, options) => {
     after((done) => common.teardown(done))
 
     it('should find other peers', (done) => {
-      nodeA.dht.findpeer(nodeB.peerId.id, (err, peer) => {
+      nodeA.dht.findpeer(nodeB.peerId.id, (err, res) => {
         expect(err).to.not.exist()
-        // TODO upgrade the answer, format is weird
-        expect(peer[0].Responses[0].ID).to.be.equal(nodeB.peerId.id)
+        expect(res.Responses[0].Addrs).to.deep.include(nodeB.peerId.addresses[0])
```
alanshaw: is `deep` necessary - the values are just strings right?

js/src/dht/findpeer.js
```diff
@@ -38,10 +38,9 @@ module.exports = (createCommon, options) => {
     after((done) => common.teardown(done))
 
     it('should find other peers', (done) => {
-      nodeA.dht.findpeer(nodeB.peerId.id, (err, peer) => {
+      nodeA.dht.findpeer(nodeB.peerId.id, (err, res) => {
```
alanshaw: Documentation or implementation for `dht.findpeer` needs to be updated - `res` is not a `PeerInfo` instance.

https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/DHT.md#dhtfindpeer

js/src/dht/findpeer.js
alanshaw: Can we please fix the object that is returned here to not have go-style uppercased first letter properties? - for consistency with the rest of the API.

js/src/dht/findprovs.js
```diff
@@ -60,7 +60,7 @@ module.exports = (createCommon, options) => {
         },
         (cidV0, cb) => nodeA.dht.findprovs(cidV0, cb),
         (provs, cb) => {
-          expect(provs.map((p) => p.id.toB58String()))
+          expect(provs.Responses.map((p) => p.ID))
```
alanshaw: Can we please fix the object that is returned here to not have go-style uppercased first letter properties? - for consistency with the rest of the API.

alanshaw: Documentation or implementation for `dht.findprovs` needs to be updated - `provs` is not an array of `PeerInfo` instances.

https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/DHT.md#dhtfindprovs

js/src/dht/findpeer.js
vasco-santos: According to http api [docs](https://docs.ipfs.io/reference/api/http/#api-v0-dht-findpeer), the `addrs` is an array of addresses. that's why I am using the `deep.include`.

Changed the uppercase letter. However, the inconsistencies will have to be solved in `ipfs-api` in [ipfs/js-ipfs-api#890](https://github.com/ipfs/js-ipfs-api/pull/890). `go-ipfs` sends the uppercase style, and `js-ipfs` (aiming to have its core implementation compliant with this interface) must send its response in lowercase. As a result, `ipfs-api` using `js-ipfs` will receive the lowercase ones. I added the logic to deal with this inconsistency in the `ipfs-api` PR, but I am not sure if it is the better approach in this case.

js/src/dht/findpeer.js
vasco-santos: updated

js/src/dht/findprovs.js
vasco-santos: updated

js/src/dht/findprovs.js
```diff
@@ -55,12 +55,12 @@ module.exports = (createCommon, options) => {
       waterfall([
         (cb) => nodeB.object.new('unixfs-dir', cb),
         (dagNode, cb) => {
-          const cidV0 = new CID(dagNode.toJSON().multihash)
+          const cidV0 = new CID(dagNode.toJSON().hash)
           nodeB.dht.provide(cidV0, (err) => cb(err, cidV0))
```
alanshaw: `object.new` now returns a CID

PR #391, opened 2018-11-10
refactor: convert bitswap tests to use async/await

alanshaw:
NOTE: the PRs for https://github.com/ipfs/js-ipfs/issues/1670 are being merged into the `feat/async-iterators` branch. Once they're all in we can merge them into master.

---

This simply refactors the bitswap tests to use async/await and updates the documentation to remove callback support and propose a better method for documenting the parameters and return values.

depends on https://github.com/ipfs/interface-ipfs-core/pull/390


SPEC/BITSWAP.md
```diff
@@ -5,60 +5,59 @@
...
+|------|------|-------------|
+| peerId | `String` | (optional) Base 58 encoded Peer ID to get the wantlist for. Default: this node's peer ID |
 
-If no `callback` is passed, a promise is returned.
+##### Returns
 
-**Example:**
+| Type | Description |
+|------|-------------|
+| `Promise<{`<br/>&nbsp;&nbsp;`Keys<Array<Object>>`<br/>`}>` | The list of CIDs wanted by the peer. Each object in the array has a single property "/" a string CID. |
```
hugomrdias: this seems weird, html inside inline code block just to add new lines and space

alanshaw: I know 😢. I would like this to be clearer when viewed as markdown but I don't know how right now. Unfortunately multi-line code blocks in a table aren't possible. If you have any ideas for how to format this in markdown so that it reads clearly when rendered then please comment!

vmx: I've recently also started to use tables for parameters (for js-ipld). But I'll move away from that again. I find tables hard to read if you just read the unrendered Markdown.

The signature is quite complicated, I actually missed that it is an object containing a single field called `Keys`. I found that clearer in the old docs.

daviddias: > I've recently also started to use tables for parameters (for js-ipld). But I'll move away from that again. I find tables hard to read if you just read the unrendered Markdown.

👍

SPEC/BITSWAP.md
```diff
@@ -5,60 +5,59 @@
...
 
-```JavaScript
-ipfs.bitswap.wantlist((err, list) => console.log(list))
+##### Example
 
-// { Keys: [{ '/': 'QmHash' }] }
+```js
+const list = await ipfs.bitswap.wantlist()
```
achingbrain: Should this return an async iterable instead of a list?

achingbrain: I think in general we should either return single values or iterators of single values, never lists of single values.

alanshaw: I agree, the wantlist could get really big so an iterator makes sense.

> never lists of single values

Do you think this should always be true? I think we should return async iterators where it makes sense to "stream" the output but I wouldn't have thought it would be necessary for something like `ipfs.bootstrap.list`, for example...but could be persuaded for the sake of consistency.

SPEC/BITSWAP.md
alanshaw: Ok, I will think up a better way to layout this information that is readable rendered or not.

SPEC/BITSWAP.md
hugomrdias: we talking iterable ? or really async iterator ?

alanshaw:
superseded by https://github.com/ipfs/interface-js-ipfs-core/pull/547

PR #395, opened 2018-11-20
fix: ipns over pubsub tests

vasco-santos:
If we have an IPNS record value in cache, we will not hit the resolver and consequently, we are not going to subscribe the topic, as it is only subscribed when we are looking for it.

According to the above, I removed the `publish` prior to the `resolve` in the tests, which lead to some changes in the structure of the test, as resolve will return an error, as a consequence of not having any record.

### DIFF
--- a/js/src/name-pubsub/cancel.js
+++ b/js/src/name-pubsub/cancel.js
@@ -2,7 +2,8 @@
 /* eslint-env mocha */
 'use strict'
 
-const series = require('async/series')
+const auto = require('async/auto')
+const PeerId = require('peer-id')
 
 const { spawnNodeWithId } = require('../utils/spawn')
 const { getDescribe, getIt, expect } = require('../utils/mocha')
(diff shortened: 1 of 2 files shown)


js/src/name-pubsub/cancel.js
```diff
@@ -63,24 +52,30 @@ module.exports = (createCommon, options) => {
...
+      const ipnsPath = `/ipns/${id}`
+
+      ipfs.name.pubsub.subs((err, res) => {
         expect(err).to.not.exist()
-        expect(res).to.exist()
-        expect(res[0]).to.eql([]) // initally empty
-        expect(res[4]).to.have.property('canceled')
-        expect(res[4].canceled).to.eql(true)
-        expect(res[5]).to.be.an('array').that.does.not.include(ipnsPath)
+        expect(res).to.eql([]) // initally empty
```
alanshaw: Since the node is shared between tests we can't assume this will be the case. We should assert that the thing you don't want to be in there is not there instead.

js/src/name-pubsub/cancel.js
```diff
@@ -63,24 +52,30 @@ module.exports = (createCommon, options) => {
...
-      series([
-        (cb) => ipfs.name.pubsub.subs(cb),
-        (cb) => ipfs.name.publish(value, { resolve: false }, cb),
-        (cb) => ipfs.name.resolve(nodeId, cb),
-        (cb) => ipfs.name.pubsub.subs(cb),
-        (cb) => ipfs.name.pubsub.cancel(ipnsPath, cb),
-        (cb) => ipfs.name.pubsub.subs(cb)
-      ], (err, res) => {
+      const id = 'QmNP1ASen5ZREtiJTtVD3jhMKhoPb1zppET1tgpjHx2NGA'
+      const ipnsPath = `/ipns/${id}`
```
alanshaw: Using a hard coded peer ID like this is fine as long as other developers don't come along and copy it and use it in their tests. Personally I'd generate one for the test.

js/src/name-pubsub/cancel.js
```diff
@@ -63,24 +52,30 @@ module.exports = (createCommon, options) => {
...
+            (cb) => ipfs.name.pubsub.subs(cb),
+            (cb) => ipfs.name.pubsub.cancel(ipnsPath, cb),
+            (cb) => ipfs.name.pubsub.subs(cb)
+          ], (err, res) => {
+            expect(err).to.not.exist()
+            expect(res).to.exist()
+            expect(res[0]).to.be.an('array').that.does.include(ipnsPath)
+            expect(res[1]).to.have.property('canceled')
+            expect(res[1].canceled).to.eql(true)
+            expect(res[2]).to.be.an('array').that.does.not.include(ipnsPath)
```
alanshaw: A note for the future, if you use `async/auto` instead you can give the results names. That way if we add another thing to the pipeline for this we don't have to update all our array indexes.

vasco-santos: Thanks! Good tip, I changed it

js/src/name-pubsub/cancel.js
```diff
@@ -63,24 +53,35 @@ module.exports = (createCommon, options) => {
...
+          expect(err).to.not.exist()
+          expect(res).to.be.an('array').that.does.not.include(ipnsPath)
+
+          ipfs.name.resolve(id, (err) => {
+            expect(err).to.exist()
+            auto({
+              subs1: (cb) => ipfs.name.pubsub.subs(cb),
+              cancel: (cb) => ipfs.name.pubsub.cancel(ipnsPath, cb),
+              subs2: (cb) => ipfs.name.pubsub.subs(cb)
+            }, 1, (err, res) => {
```
alanshaw: With auto you need to specify your dependencies otherwise this will all run in parallel:

````js
            auto({
              subs1: (cb) => ipfs.name.pubsub.subs(cb),
              cancel: ['subs1', (_, cb) => ipfs.name.pubsub.cancel(ipnsPath, cb)],
              subs2: ['cancel', (_, cb) => ipfs.name.pubsub.subs(cb)]
            } /* ... */)
```

PR #400, opened 2018-11-27
chore: increase ping timeouts

alanshaw:
This takes around 15s locally and can take significantly longer in CI.

screenshot 2018-11-27 at 13 37 41

### DIFF
--- a/js/src/ping/ping-pull-stream.js
+++ b/js/src/ping/ping-pull-stream.js
@@ -14,7 +14,7 @@ module.exports = (createCommon, options) => {
   const common = createCommon()
 
   describe('.pingPullStream', function () {
-    this.timeout(60 * 1000)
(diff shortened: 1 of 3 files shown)


hugomrdias:
shouldn't we try to improve the code instead of just increasing the mocha timeout ? This shouldn't take this long right 2 * 60 * 1000?

alanshaw:
The screenshot above is from js-ipfs-api talking to a go-ipfs node. Most of the time it takes ~15s but occasionally it takes >1min in CI. I'll open an issue on the go-ipfs repo.

alanshaw:
https://github.com/ipfs/go-ipfs/issues/5799

alanshaw:
@hugomrdias I've added comments with link to the issue. I'm going to merge this so that our ping tests become more reliable and we can revisit when the issue is resolved.

PR #401, opened 2018-11-28
docs: add mfs stream ls methods

achingbrain:
The only `mfs.ls` method at the moment buffers the output into an array before returning it to the user. This PR adds two new methods, `lsPullStream` and `lsReadableStream` to allow the user to either buffer the output themselves (in case they need sorting, etc) or just stream it on to an output of some sort.

### DIFF
--- a/SPEC/FILES.md
+++ b/SPEC/FILES.md
@@ -22,6 +22,8 @@
   - [files.cp](#filescp)
   - [files.flush](#filesflush)
   - [files.ls](#filesls)
+  - [files.lsReadableStream](#fileslsreadablestream)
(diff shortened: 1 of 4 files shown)


daviddias:
I'm with @hugomrdias and @alanshaw

achingbrain:
It think it's better for them to be consistent with the existing APIs, then we can break them all in one go down the line.

alanshaw:
Can we get some tests in here? Can we support it in `ipfs-http-client` with the new streaming ls in go-ipfs?

achingbrain:
Tests are added. ipfs/js-ipfs-http-client#903 is the other part of this.

PR #407, opened 2018-12-05
fix: allow only by object

alanshaw:
The only option will now accept an object for onlying a specific test.

Previously you could only do this to only run a specific test:

```js
{
  only: ['should test a thing']
}
```

Now you can do this:

```js
{
  only: [{
    name: 'should test a thing',
    reason: 'because development' // optional!
  }]
}
```

License: MIT
Signed-off-by: Alan Shaw <alan.shaw@protocol.ai>

### DIFF
--- a/js/src/utils/mocha.js
+++ b/js/src/utils/mocha.js
@@ -59,7 +59,14 @@ function getIt (config) {
     }
 
     if (Array.isArray(config.only)) {
-      if (config.only.includes(name)) return it.only(name, impl) // eslint-disable-line
+      const only = config.only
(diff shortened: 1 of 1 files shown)


achingbrain:
Is this in a README somewhere? I find myself rummaging around in my memory/the codebase a lot for this sort of thing.

alanshaw:
yep its all in the README! https://github.com/ipfs/interface-ipfs-core/blob/master/README.md#running-only-some-tests

alanshaw:
Assuming this is ok....am merge. Shout if this is not what you wanted!

PR #413, opened 2018-12-09
test: add CID version agnostic tests

alanshaw:


js/src/block/get.js
```diff
@@ -4,6 +4,7 @@
 const multihash = require('multihashes')
 const CID = require('cids')
 const auto = require('async/auto')
+const crypto = require('crypto')
```
daviddias: Do the bytes need to be so random that it requires to add the crypto module (and respective browser polyfill?)

alanshaw: No, and it's probably faster to just use `Date.now()` or `Math.random()`. I've removed all instances of `crypto`. There's one test that generates lots of random data to ensure it spans multiple dag nodes and I've changed that test to use [`randombytes`](https://www.npmjs.com/package/randombytes) so we don't pull in the whole of `crypto` in the browser.

alanshaw:
Assuming this PR is gtg now @daviddias - please shout if you have further changes and I'll submit another PR!

PR #419, opened 2018-12-14
support bundle size PRs

hugomrdias:
WIP


alanshaw (changes requested):
Please could you also update the docs where we reference big.js?

alanshaw (approved):

PR #420, opened 2018-12-14
test: add test for ls by IPFS path

alanshaw:
for https://github.com/ipfs/js-ipfs-http-client/pull/914


vasco-santos:
Linting errors, as exceed the max number of callbacks. Can you just add the `eslint` ignore?

PR #426, opened 2019-01-08
files.ls options name: long to l

alx696:
`l` work, `long` not.

go-ipfs v0.4.18

### DIFF
--- a/SPEC/FILES.md
+++ b/SPEC/FILES.md
@@ -1090,7 +1090,7 @@ Where:
 
 - `path` is an optional string to show listing for (default: `/`)
 - `options` is an optional Object that might contain the following keys:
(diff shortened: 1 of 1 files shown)


achingbrain:
go-ipfs should support -l and —long - would you like to try adding support for it?

Ref: https://github.com/ipfs/go-ipfs/issues/5026#issuecomment-392477565

alx696:
> go-ipfs should support -l and —long - would you like to try adding support for it?
> 
> Ref: [ipfs/go-ipfs#5026 (comment)](https://github.com/ipfs/go-ipfs/issues/5026#issuecomment-392477565)

Sorry! It is js-ipfs-http-client problem. Not go-ipfs.

`ipfs.files.ls(path, {long: true})` not work.

Issue #436, opened 2019-02-06
adding unsubscribe(eventName) 


Isan-Rivkin:
Hi, currently `unsubscribe` method must take a reference to the `handler` and it forces me to keep a reference to it while it may not be ideal for some implementations. 
Maybe `unsubscribe(eventName)` could be provided based on [EventEmitter.removeAllListeners(eventName)](https://nodejs.org/api/events.html#events_emitter_removealllisteners_eventname) and how `floodsub` is implemented (inherits EventEmitter). 
I don't mind creating PR for this, just wonder if there's an issue with that first,

alanshaw:
I think that's a good suggestion. We'd need a few PRs to make this happen:

1. PR to add a test in this repo and update the documentation
2. PR to `js-libp2p` https://github.com/libp2p/js-libp2p/blob/4ed5c039fc8583b62cb1e41284a025f5d8d258bc/src/pubsub.js#L41 to call `removeAllListeners` instead if `handler == null`
2. PR to `ipfs-http-client` to do the same thing https://github.com/ipfs/js-ipfs-http-client/blob/bfd71c2bd9a467b166c85eaa7dd5d7fb94e4dfa4/src/pubsub.js#L80

@jacobheun @vasco-santos is that cool with you?

jacobheun:
> @jacobheun @vasco-santos is that cool with you?

Yes, this sounds great.

Isan-Rivkin:
moved to PR #437

PR #437, opened 2019-02-06
pubsub.unsubscribe without handler reference

Isan-Rivkin:
Hi, continuing issue #436  here as a PR preparing. (Not ready yet)
I created a PR both on [js-ipfs-http-client](https://github.com/ipfs/js-ipfs-http-client/pull/940) and [js-libp2p](https://github.com/libp2p/js-libp2p/pull/321) as @alanshaw suggested.
If everything ok I will add tests to this repo as well. 

Let me know what you think.

### DIFF
--- a/SPEC/PUBSUB.md
+++ b/SPEC/PUBSUB.md
@@ -51,6 +51,8 @@ A great source of [examples][] can be found in the tests for this API.
 If no `callback` is passed, a [promise][] is returned.
 
 This works like `EventEmitter.removeListener`, as that only the `handler` passed to a `subscribe` call before is removed from listening. The underlying subscription will only be canceled once all listeners for a topic have been removed.
+The other options is to use this method with only `topic` as input, without the need to keep a reference to the `handler` function. 
+This works like `EventEmitter.remoteAllListeners`, removes all listeners from the given topic. 
 
 **Example:**
 
@@ -74,6 +76,25 @@ ipfs.pubsub.subscribe(topic, receiveMsg, (err) => {
 
       console.log(`unsubscribed from ${topic}`)
     })
+  }, 1000)
+})
+```
+
+Or removing all listeners: 
+```JavaScript
+const topic = 'fruit-of-the-day'
+const receiveMsg = (msg) => console.log(msg.toString())
+
+ipfs.pubsub.subscribe(topic, receiveMsg, (err) => {
+  if (err) {
+    return console.error(`failed to subscribe to ${topic}`, err)
+  }
+
+  console.log(`subscribed to ${topic}`)
+
+  setTimeout(() => {
+    // Will unsubscribe ALL handlers for the given topic
+    ipfs.pubsub.unsubscribe(topic);
   }, 1000)
 })
 ```
--- a/SPEC/PUBSUB.md
+++ b/SPEC/PUBSUB.md
@@ -48,11 +48,17 @@ A great source of [examples][] can be found in the tests for this API.
 - `handler: (msg) => {}` - The handler to remove.
 - `callback: (Error) => {}` (Optional) Called once the unsubscribe is done.
 
-If a **handler is passed** and no `callback` is passed, a [promise][] is returned.
+If no `callback` is passed, a [promise][] is returned.
 
-If the `topic` and `handler` are provided, the `handler` will no longer receive updates for the `topic`. This behaves like `EventEmitter.removeListener`. If the `handler` is not equivalent to the `handler` provided on `subscribe`, no action will be taken.
+If the `topic` and `handler` are provided, the `handler` will no longer receive updates for the `topic`. This behaves like [EventEmitter.removeListener](https://nodejs.org/dist/latest/docs/api/events.html#events_emitter_removelistener_eventname_listener). If the `handler` is not equivalent to the `handler` provided on `subscribe`, no action will be taken.
 
-If **only** the `topic` param is provided, unsubscribe will remove **all** handlers for the `topic`. This behaves like `EventEmitter.remoteAllListeners`. Use this if you would like to no longer receive any updates for the `topic`.
+If **only** the `topic` param is provided, unsubscribe will remove **all** handlers for the `topic`. This behaves like [EventEmitter.removeAllListeners](https://nodejs.org/dist/latest/docs/api/events.html#events_emitter_removealllisteners_eventname). Use this if you would like to no longer receive any updates for the `topic`.
+
+**WARNING:** Unsubscribe is an async operation, but removing **all** handlers for a topic can only be done using the Promises API (due to the difficulty in distinguishing between a "handler" and a "callback" - they are both functions). If you _need_ to know when unsubscribe has completed you must use `await` or `.then` on the return value from 
+
+```JavaScript
+ipfs.pubsub.unsubscribe('topic')
+```
 
 **Example:**
 
@@ -85,10 +91,10 @@ Or removing all listeners:
 const topic = 'fruit-of-the-day'
 const receiveMsg = (msg) => console.log(msg.toString())
 
-ipfs.pubsub.subscribe(topic, receiveMsg).then(() => {
-  // Will unsubscribe ALL handlers for the given topic
-  ipfs.pubsub.unsubscribe(topic);
-})
+await ipfs.pubsub.subscribe(topic, receiveMsg);
+
+// Will unsubscribe ALL handlers for the given topic
+await ipfs.pubsub.unsubscribe(topic);
 ```
 
 A great source of [examples][] can be found in the tests for this API.
--- a/SPEC/PUBSUB.md
+++ b/SPEC/PUBSUB.md
@@ -48,7 +48,7 @@ A great source of [examples][] can be found in the tests for this API.
 - `handler: (msg) => {}` - The handler to remove.
 - `callback: (Error) => {}` (Optional) Called once the unsubscribe is done.
 
-If no `callback` is passed, a [promise][] is returned.
+If a **handler is passed** and no `callback` is passed, a [promise][] is returned.
 
 If the `topic` and `handler` are provided, the `handler` will no longer receive updates for the `topic`. This behaves like `EventEmitter.removeListener`. If the `handler` is not equivalent to the `handler` provided on `subscribe`, no action will be taken.
 
--- a/SPEC/PUBSUB.md
+++ b/SPEC/PUBSUB.md
@@ -85,17 +85,9 @@ Or removing all listeners:
 const topic = 'fruit-of-the-day'
 const receiveMsg = (msg) => console.log(msg.toString())
 
-ipfs.pubsub.subscribe(topic, receiveMsg, (err) => {
-  if (err) {
-    return console.error(`failed to subscribe to ${topic}`, err)
-  }
-
-  console.log(`subscribed to ${topic}`)
-
-  setTimeout(() => {
-    // Will unsubscribe ALL handlers for the given topic
-    ipfs.pubsub.unsubscribe(topic);
-  }, 1000)
+ipfs.pubsub.subscribe(topic, receiveMsg).then(() => {
+  // Will unsubscribe ALL handlers for the given topic
+  ipfs.pubsub.unsubscribe(topic);
 })
 ```
 
--- a/SPEC/PUBSUB.md
+++ b/SPEC/PUBSUB.md
@@ -50,9 +50,9 @@ A great source of [examples][] can be found in the tests for this API.
 
 If no `callback` is passed, a [promise][] is returned.
 
-This works like `EventEmitter.removeListener`, as that only the `handler` passed to a `subscribe` call before is removed from listening. The underlying subscription will only be canceled once all listeners for a topic have been removed.
-The other options is to use this method with only `topic` as input, without the need to keep a reference to the `handler` function. 
-This works like `EventEmitter.remoteAllListeners`, removes all listeners from the given topic. 
+If the `topic` and `handler` are provided, the `handler` will no longer receive updates for the `topic`. This behaves like `EventEmitter.removeListener`. If the `handler` is not equivalent to the `handler` provided on `subscribe`, no action will be taken.
+
+If **only** the `topic` param is provided, unsubscribe will remove **all** handlers for the `topic`. This behaves like `EventEmitter.remoteAllListeners`. Use this if you would like to no longer receive any updates for the `topic`.
 
 **Example:**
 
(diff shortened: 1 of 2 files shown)


alanshaw:
Ah, I thought you were suggesting that calling `pubsub.unsubscribe` without a handler would unsubscribe all. e.g. `pubsub.unsubscribe('mytopic', err => {})`

jacobheun:
> Ah, I thought you were suggesting that calling `pubsub.unsubscribe` without a handler would unsubscribe all. e.g. `pubsub.unsubscribe('mytopic', err => {})`

Me too. I would prefer to see `unsubscribe` just be overloaded and the documentation for it updated.

Isan-Rivkin:
@alanshaw @jacobheun I thought so too, but the issue is that `unsubscribe` takes 3 parameters 
- topic
- handler(msg)=>{} 
- callback(err)=>{}
`callback` is optional, so if you do overload and the caller passes 2 arguments how will you know which is is it? handler or callback? Since both of them are optional now. 
I would love to overload it if you have an idea and I'm missing something :-)

jacobheun:
> * `callback` is optional, so if you do overload and the caller passes 2 arguments how will you know which one is it? `handler` or `callback`?

Bleh, that's right. My initial thought is to use this as an opportunity to change the pubsub system over to async only, but having different system in different states would probably be unpleasant, although they technically already support promises. @alanshaw what do you think? 

`unsubscribeAll` is probably the easiest approach if we don't change to async only. Otherwise callback would need to become not optional.

Isan-Rivkin:
@jacobheun I can either stay with `unsubscribeAll` or make `callback` not optional. 
The all-async options will probably take a little longer 0_0

alanshaw:
Ah, good point.

I think this might have to wait for https://github.com/ipfs/js-ipfs/issues/1670.

What do you think @jacobheun - I'm hesitant to add an additional API method for this...

jacobheun:
I think doing this as part of the async iterators work is the better way to go. @alanshaw how do you feel about switching over individual subsystems (perhaps this is a conversation to have with the larger group)? We currently support promises for pubsub, so starting to remove support for callbacks system by system, and communicating those out, might be a reasonable approach. It would let us iterate on the APIs, and also allow users to update their components along with us, rather than large, significant refactors. Migration guides for each release would also be a bit easier to produce and consume.

alanshaw:
In theory I'm pro this - it makes the process easier for us, but I think in this case we shouldn't do it. It makes it really difficult for users to know which release has which API and we also don't have an easy way to switch release versions in our API docs so it'll get confusing and we'll get a ton of issues saying "never calls callback" for people who upgrade, see callbacks working with some API methods but not others.

Isan-Rivkin:
Just to clarify, this is on hold?

jacobheun:
I think there is another option to get us the unsubscribeAll functionality without waiting. If `unsubscribe` only receives a single param, perform an unsubscribe all. Since callback is optional anyway, we could just say that if you want to unsubscribe all handlers for a topic, you **must only** include the topic name as a parameter.

```js
// Will unsubscribe the handler for the given topic and call the callback if it's provided
ipfs.pubsub.unsubscribe(topic, handler, [callback])
```

```js
// Will unsubscribe ALL handlers for the given topic
ipfs.pubsub.unsubscribe(topic)
```

Thoughts?

Isan-Rivkin:
@jacobheun under current constraints it seems like a fair solution, im ok with adjusting the PR let me know 👍

alanshaw:
SGTM

jacobheun:
@Isan-Rivkin yes, let's move forward with the new approach.

Isan-Rivkin:
this got accidentally closed because I merged this repo into my fork and they were aligned.

Isan-Rivkin:
@jacobheun done, here and  both on [js-ipfs-http-client](https://github.com/ipfs/js-ipfs-http-client/pull/940) and [js-libp2p](https://github.com/libp2p/js-libp2p/pull/321) 
Not sure what to do about the codecov in `js-ipfs-http-client`?

src/pubsub/unsubscribe.js
```diff
@@ -57,5 +57,21 @@ module.exports = (createCommon, options) => {
         )
       })
     })
+    it('should subscribe and unsubscribe with no handler', (done) => {
+      const someTopic = getTopic()
+      const handler = (msg) => {}
+      ipfs.pubsub.subscribe(someTopic, handler, (err) => {
```
jacobheun: It would be good to add more than one handler for `someTopic` so we're verifying it is indeed removing everything.

SPEC/PUBSUB.md
```diff
@@ -51,6 +51,8 @@ A great source of [examples][] can be found in the tests for this API.
 If no `callback` is passed, a [promise][] is returned.
 
 This works like `EventEmitter.removeListener`, as that only the `handler` passed to a `subscribe` call before is removed from listening. The underlying subscription will only be canceled once all listeners for a topic have been removed.
+The other options is to use this method with only `topic` as input, without the need to keep a reference to the `handler` function. 
+This works like `EventEmitter.remoteAllListeners`, removes all listeners from the given topic.
```
jacobheun: I think it might be good to change how this section is worded (lines 53-55). I think the following might be a bit clearer:
```markdown
If the `topic` and `handler` are provided, the `handler` will no longer receive updates for the `topic`. This behaves like `EventEmitter.removeListener`. If the `handler` is not equivalent to the `handler` provided on `subscribe`, no action will be taken.

If **only** the `topic` param is provided, unsubscribe will remove **all** handlers for the `topic`. This behaves like `EventEmitter.remoteAllListeners`. Use this if you would like to no longer receive any updates for the `topic`.
```

jacobheun:
Added some minor suggestions. Codecov in the http client will get resolved once this is released, as it uses the test suite from here.

src/pubsub/unsubscribe.js
Isan-Rivkin: done

SPEC/PUBSUB.md
Isan-Rivkin: done

Isan-Rivkin:
@jacobheun Thanks for the responsiveness! I think resolved.

Isan-Rivkin:
@alanshaw kindly pinging you for review :)

SPEC/PUBSUB.md
```diff
@@ -50,7 +50,9 @@ A great source of [examples][] can be found in the tests for this API.
 
 If no `callback` is passed, a [promise][] is returned.
 
-This works like `EventEmitter.removeListener`, as that only the `handler` passed to a `subscribe` call before is removed from listening. The underlying subscription will only be canceled once all listeners for a topic have been removed.
+If the `topic` and `handler` are provided, the `handler` will no longer receive updates for the `topic`. This behaves like `EventEmitter.removeListener`. If the `handler` is not equivalent to the `handler` provided on `subscribe`, no action will be taken.
+
+If **only** the `topic` param is provided, unsubscribe will remove **all** handlers for the `topic`. This behaves like `EventEmitter.remoteAllListeners`. Use this if you would like to no longer receive any updates for the `topic`.
```
alanshaw: I think we need an extra note here to reinforce that it can ONLY be used with the Promise API. I reckon people will still think they can pass topic+callback.

SPEC/PUBSUB.md
```diff
@@ -78,6 +80,25 @@ ipfs.pubsub.subscribe(topic, receiveMsg, (err) => {
...
+  if (err) {
+    return console.error(`failed to subscribe to ${topic}`, err)
+  }
+
+  console.log(`subscribed to ${topic}`)
+
+  setTimeout(() => {
+    // Will unsubscribe ALL handlers for the given topic
+    ipfs.pubsub.unsubscribe(topic);
+  }, 1000)
```
alanshaw: Lets use `async/await` in this example, we shouldn't encourage people to use `setTimeout`.

src/pubsub/unsubscribe.js
```diff
@@ -57,5 +57,26 @@ module.exports = (createCommon, options) => {
...
+        setTimeout(() => {
+          // Assert unsubscribe worked
+          ipfs.pubsub.ls((err, topics) => {
+            expect(err).to.not.exist()
+            expect(topics).to.eql([])
+            done()
+          })
+        }, 500)
+      })
+    })
```
alanshaw: Can we please use `async/await` or promises directly in this test? 500ms delay isn't guarateed to be enough time to wait.

Isan-Rivkin: done

SPEC/PUBSUB.md
Isan-Rivkin: This is not required or something I was just following up on [previous example](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/PUBSUB.md#pubsubunsubscribe), 
anyway I have changed it because [subscribe](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/PUBSUB.md#ipfspubsubunsubscribetopic-handler-callback) returns a Promise as you mentioned. 
I specifically didn't use `async/await` but instead `.then()` because JS requires using `await` inside `async` method keyword so IDK didn't feel right, let me know if you want it to be changed.

SPEC/PUBSUB.md
Isan-Rivkin: Added extra note on the promise

Isan-Rivkin:
@alanshaw done, thanks! let me know if anything should be changed.

SPEC/PUBSUB.md
```diff
@@ -48,9 +48,11 @@ A great source of [examples][] can be found in the tests for this API.
 - `handler: (msg) => {}` - The handler to remove.
 - `callback: (Error) => {}` (Optional) Called once the unsubscribe is done.
 
-If no `callback` is passed, a [promise][] is returned.
+If a **handler is passed** and no `callback` is passed, a [promise][] is returned.
```
alanshaw: I think this was ok as it was before...

SPEC/PUBSUB.md
```diff
@@ -48,9 +48,11 @@ A great source of [examples][] can be found in the tests for this API.
...
+
+If the `topic` and `handler` are provided, the `handler` will no longer receive updates for the `topic`. This behaves like `EventEmitter.removeListener`. If the `handler` is not equivalent to the `handler` provided on `subscribe`, no action will be taken.
```
alanshaw: Can we get a link to https://nodejs.org/dist/latest/docs/api/events.html#events_emitter_removelistener_eventname_listener here?

SPEC/PUBSUB.md
```diff
@@ -48,9 +48,11 @@ A great source of [examples][] can be found in the tests for this API.
...
 
-This works like `EventEmitter.removeListener`, as that only the `handler` passed to a `subscribe` call before is removed from listening. The underlying subscription will only be canceled once all listeners for a topic have been removed.
+If **only** the `topic` param is provided, unsubscribe will remove **all** handlers for the `topic`. This behaves like `EventEmitter.remoteAllListeners`. Use this if you would like to no longer receive any updates for the `topic`.
```
alanshaw: Typo "remoteAllListeners" => "removeAllListeners".

alanshaw: Can we get a link to https://nodejs.org/dist/latest/docs/api/events.html#events_emitter_removealllisteners_eventname here also?

alanshaw: I was expecting something like this at the end of this paragraph:

> WARNING: Unsubscribe is an async operation, but removing **all** handlers for a topic can only be done using the Promises API (due to the difficulty in distinguishing between a "handler" and a "callback" - they are both functions). If you _need_ to know when unsubscribe has completed you must use `await` or `.then` on the return value from `ipfs.pubsub.unsubscribe('topic')`.

SPEC/PUBSUB.md
```diff
@@ -78,6 +80,17 @@ ipfs.pubsub.subscribe(topic, receiveMsg, (err) => {
...
+Or removing all listeners: 
+```JavaScript
+const topic = 'fruit-of-the-day'
+const receiveMsg = (msg) => console.log(msg.toString())
+
+ipfs.pubsub.subscribe(topic, receiveMsg).then(() => {
+  // Will unsubscribe ALL handlers for the given topic
+  ipfs.pubsub.unsubscribe(topic);
+})
+```
```
alanshaw: I was looking for something like this:

```JavaScript
const topic = 'fruit-of-the-day'
const receiveMsg = (msg) => console.log(msg.toString())

await ipfs.pubsub.subscribe(topic, receiveMsg)

// Will unsubscribe ALL handlers for the given topic
await ipfs.pubsub.unsubscribe(topic)
```

src/pubsub/unsubscribe.js
```diff
@@ -57,5 +57,24 @@ module.exports = (createCommon, options) => {
...
+        expect(err).to.not.exist()
+        ipfs.pubsub.unsubscribe(someTopic)
+        // Assert unsubscribe worked
+        ipfs.pubsub.ls((err, topics) => {
+          expect(err).to.not.exist()
+          expect(topics).to.eql([])
+          done()
+        })
+      })
+    })
```
alanshaw: We need to `await` on the unsubscribe, it won't necessarily be complete until after the promise resolves:

```js
    it('should subscribe 10 handlers and unsunscribe once with no reference to the handlers', async () => {
      const count = 10
      const someTopic = getTopic()

      for (let i = 0; i < count; i++) {
        await ipfs.pubsub.subscribe(someTopic, (msg) => {})
      }

      await ipfs.pubsub.unsubscribe(someTopic)
      const topics =  await ipfs.pubsub.ls()
      expect(topics).to.eql([])
    })
```

Isan-Rivkin:
@alanshaw thanks for that review and that you provide examples! changed what you asked, let me know what you think

src/pubsub/unsubscribe.js
Isan-Rivkin: added, thanks! didn't notice unsubscribe is part of the async api

SPEC/PUBSUB.md
Isan-Rivkin: done

Isan-Rivkin: done

SPEC/PUBSUB.md
Isan-Rivkin: done

SPEC/PUBSUB.md
Isan-Rivkin: done

SPEC/PUBSUB.md
Isan-Rivkin: fixed

SPEC/PUBSUB.md
Isan-Rivkin: done

PR #440, opened 2019-02-19
fix: don't expect ipfs to preserve a leading slash

Stebalien:
Go-ipfs now normalizes paths before adding files. This will:

* remove any leading slashes.
* remove any `/./` components.
* normalize any `/../` components.

That breaks this test which assumes that paths are returned _exactly_ as specified.

### DIFF
--- a/src/miscellaneous/resolve.js
+++ b/src/miscellaneous/resolve.js
@@ -71,7 +71,7 @@ module.exports = (createCommon, options) => {
 
     // Test resolve turns /ipfs/QmRootHash/path/to/file into /ipfs/QmFileHash
     it('should resolve an IPFS path link', (done) => {
-      const path = '/path/to/testfile.txt'
+      const path = 'path/to/testfile.txt'
       const content = loadFixture('test/fixtures/testfile.txt', 'interface-ipfs-core')
 
       ipfs.add([{ path, content }], { wrapWithDirectory: true }, (err, res) => {
@@ -80,7 +80,7 @@ module.exports = (createCommon, options) => {
         const rootHash = res.find(r => r.path === '').hash
         const fileHash = res.find(r => r.path === path).hash
 
-        ipfs.resolve(`/ipfs/${rootHash}${path}`, (err, path) => {
+        ipfs.resolve(`/ipfs/${rootHash}/${path}`, (err, path) => {
           expect(err).to.not.exist()
           expect(path).to.equal(`/ipfs/${fileHash}`)
           done()


alanshaw:
I'm glad these are now normalized, we'll have to submit a PR to JS IPFS to implement this. It needs some tests to ensure the normalisation happens.

Separately, I think this change breaks the path given to resolve later in the test: `/ipfs/${rootHash}${path}`, I think we'd end up with `/ipfs/QmHashpath/to/testfile.txt`.

alanshaw:
@Stebalien would you mind pointing me in the direction of the PR(s) that implemented this? I just want to make sure we do the same thing...For instance, I'd like to know what happens with paths that try to traverse above root e.g. '../foo' => 'foo' or '../foo' => error? Also windows style paths i.e. backslashes and `C:\\`.

alanshaw:
See https://github.com/ipfs/go-ipfs-files/pull/6 and https://golang.org/pkg/path/filepath/#Clean

> (Note: we might want to consider rejecting .. instead of normalizing)

Stebalien:
> Separately, I think this change breaks the path given to resolve later in the test: /ipfs/${rootHash}${path}, I think we'd end up with /ipfs/QmHashpath/to/testfile.txt.

(got to get better about actually running tests)

---

This changed (incidentally) in https://github.com/ipfs/go-ipfs-files/pull/6. Normalized with  https://golang.org/pkg/path/filepath/#Clean.

Actually, I think we're going to need to think through this a bit:

* `Clean` uses platform-specific separators where we should just be using `/` at this point.
* `Clean` normalizes `..` while we should probably reject it.

I think we need a custom clean function. I'll make a patch for go-ipfs.

Stebalien:
> (got to get better about actually running tests)

Ah, that's why I didn't. Didn't want to learn how to replace npm deps.

alanshaw:
> Actually, I think we're going to need to think through this a bit. `Clean` normalizes `..` while we should probably reject it. Thoughts?

Do you mean `..` when attempting to navigate up from root or `..` anywhere in the path?

For the former, I don't have strong feelings on it. It's pretty common to _not error_ when attempting to navigate up from root, go-lang `Clean` being the case in point. But also hyperlinks on the internet, `path.resolve` in Node.js and `cd / && cd ../ && pwd` works. That said, a traversal up from root is either user error or malicious so I'd be ok to error in this situation. On balance I think I'm slightly more inclined to resolve `/..` to `/` without error.

For the latter, I guess it's just convenient to allow this when adding and it's easy to do the resolve. People will just expect it to work so why get in the way?

I noticed that `Clean` doesn't specify anything about '\'...do you think we should replace `\` with `/` and trim `/[a-zA-Z]:\\/` from the start?

Stebalien:
> Do you mean .. when attempting to navigate up from root or .. anywhere in the path?

Both? The request body feels like a declarative filesystem layout. For example, I I'd expect `tar` to reject tar files with `..` in paths.

But I don't really fell all that strong about this.

PR #444, opened 2019-02-27
refactor: tests for CIDv1 base32

alanshaw:


lidel:
@Stebalien I believe this PR may be a good place to share what cidvb32 tests should be added to the interop suite to account for blockstore/dht instricacies.

Do you have anything specific in mind, that we should always check?

PR #454, opened 2019-04-04
fix: swarm addrs test

alanshaw:
If the common factory creates a node with no bootstrap nodes then getting a list of swarm addresses could be empty (or could also be empty anyway if the connection to the bootstrap nodes takes a long time). This PR spawns and connects another IPFS node to ensure that there should be swarm addresses available.

### DIFF
--- a/src/swarm/addrs.js
+++ b/src/swarm/addrs.js
@@ -1,7 +1,9 @@
 /* eslint-env mocha */
 'use strict'
 
+const PeerInfo = require('peer-info')
 const { getDescribe, getIt, expect } = require('../utils/mocha')
+const { spawnNodesWithId } = require('../utils/spawn')
 
 module.exports = (createCommon, options) => {
   const describe = getDescribe(options)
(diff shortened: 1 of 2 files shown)


src/swarm/addrs.js
```diff
@@ -20,28 +22,30 @@ module.exports = (createCommon, options) => {
...
+          ipfsA.swarm.connect(ipfsB.peerId.addresses[0], done)
         })
       })
     })
 
     after((done) => common.teardown(done))
 
     it('should get a list of node addresses', (done) => {
-      ipfs.swarm.addrs((err, multiaddrs) => {
+      ipfsA.swarm.addrs((err, multiaddrs) => {
```
vasco-santos: Should we change `multiaddrs` to `peers` or `peersInfo`?

I was not understanding the `forEach` validation, but after getting into the code, I noticed that in fact `ipfs.swarm.addrs()` returns an array of `PeerInfo`. This way, calling this `multiaddrs` seems confusing. I think we should also update in the [SPEC](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/SWARM.md#ipfsswarmaddrscallback)

alanshaw: I changed to `peerInfos` as there's already precedent for that name in `swarm.peers` docs.

PR #458, opened 2019-04-22
feat: add support to resolve dns to ipns

hugomrdias:
This PR also normalizes `ipns name resolve` return value and offline options, to integrate better with http client and ipfs core.

### DIFF
--- a/src/name/publish.js
+++ b/src/name/publish.js
@@ -12,7 +12,7 @@ module.exports = (createCommon, options) => {
   const it = getIt(options)
   const common = createCommon()
 
-  describe('.name.publish', function () {
+  describe('.name.publish offline', function () {
     const keyName = hat()
     let ipfs
     let nodeId
@@ -43,7 +43,7 @@ module.exports = (createCommon, options) => {
 
       const value = fixture.cid
 
-      ipfs.name.publish(value, (err, res) => {
+      ipfs.name.publish(value, { 'allow-offline': true }, (err, res) => {
         expect(err).to.not.exist()
         expect(res).to.exist()
         expect(res.name).to.equal(nodeId)
@@ -62,7 +62,8 @@ module.exports = (createCommon, options) => {
         resolve: false,
         lifetime: '1m',
         ttl: '10s',
-        key: 'self'
+        key: 'self',
+        'allow-offline': true
       }
 
       ipfs.name.publish(value, options, (err, res) => {
@@ -83,7 +84,8 @@ module.exports = (createCommon, options) => {
         resolve: false,
         lifetime: '24h',
         ttl: '10s',
-        key: keyName
+        key: keyName,
+        'allow-offline': true
       }
 
       ipfs.key.gen(keyName, { type: 'rsa', size: 2048 }, function (err, key) {
--- a/src/name/publish.js
+++ b/src/name/publish.js
@@ -53,6 +53,13 @@ module.exports = (createCommon, options) => {
       })
     })
 
+    it('should publish correctly with the lifetime option and resolve', async () => {
+      const [{ path }] = await ipfs.add(Buffer.from('should publish correctly with the lifetime option and resolve'))
+      await ipfs.name.publish(path, { 'allow-offline': true, resolve: false, lifetime: '2h' })
+
+      return expect(ipfs.name.resolve(`/ipns/${nodeId}`)).to.become(`/ipfs/${path}`)
+    })
+
     it('should publish correctly when the file was not added but resolve is disabled', function (done) {
       this.timeout(50 * 1000)
 
--- a/src/name/publish.js
+++ b/src/name/publish.js
@@ -57,7 +57,7 @@ module.exports = (createCommon, options) => {
       const [{ path }] = await ipfs.add(Buffer.from('should publish correctly with the lifetime option and resolve'))
       await ipfs.name.publish(path, { 'allow-offline': true, resolve: false, lifetime: '2h' })
 
-      return expect(ipfs.name.resolve(`/ipns/${nodeId}`)).to.become(`/ipfs/${path}`)
+      return expect(await ipfs.name.resolve(`/ipns/${nodeId}`)).to.eq(`/ipfs/${path}`)
     })
 
     it('should publish correctly when the file was not added but resolve is disabled', function (done) {
--- a/src/name/resolve.js
+++ b/src/name/resolve.js
@@ -20,10 +20,6 @@ module.exports = (createCommon, options) => {
     let keyId
 
     before(function (done) {
-      // CI takes longer to instantiate the daemon, so we need to increase the
-      // timeout for the before step
-      this.timeout(60 * 1000)
-
       common.setup((err, factory) => {
         expect(err).to.not.exist()
 
@@ -41,8 +37,6 @@ module.exports = (createCommon, options) => {
     after((done) => common.teardown(done))
 
     it('should resolve a record with the default params after a publish', function (done) {
-      this.timeout(50 * 1000)
-
       const value = fixture.cid
 
       ipfs.name.publish(value, { 'allow-offline': true }, (err, res) => {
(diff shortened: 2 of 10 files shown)


src/name/resolve.js
```diff
@@ -45,14 +45,14 @@ module.exports = (createCommon, options) => {
...
-      ipfs.name.publish(value, (err, res) => {
+      ipfs.name.publish(value, { 'allow-offline': true }, (err, res) => {
         expect(err).to.not.exist()
         expect(res).to.exist()
 
         ipfs.name.resolve(nodeId, (err, res) => {
           expect(err).to.not.exist()
           expect(res).to.exist()
-          expect(res.path).to.equal(`/ipfs/${value}`)
+          expect(res).to.equal(`/ipfs/${value}`)
```
lidel: Was it always failing before (when `.path` was expected), or is this change introduced by other PR?

src/name/resolve.js
```diff
@@ -62,28 +62,31 @@ module.exports = (createCommon, options) => {
...
-        lifetime: '1ms',
+        lifetime: '100ms',
         ttl: '10s',
-        key: 'self'
+        key: 'self',
+        'allow-offline': true
       }
 
-      ipfs.name.publish(value, publishOptions, (err, res) => {
+      ipfs.add(Buffer.from('should not get the entry if its validity time expired'), (err, r) => {
```
vasco-santos: Can you add a comment explaining why we do this new `ipfs.add`? 

As we intend to publish a record with less lifetime than the one published on the previous test, `go-ipfs` will return an error for this. I am also thinking if we should just do the `beforeEach` before.

src/name/publish.js
```diff
@@ -43,7 +43,7 @@ module.exports = (createCommon, options) => {
 
       const value = fixture.cid
 
-      ipfs.name.publish(value, (err, res) => {
+      ipfs.name.publish(value, { 'allow-offline': true }, (err, res) => {
```
alanshaw: Camel case please - `allowOffline`.

src/name/resolve.js
alanshaw: Minor, but you could remove `expect(res).to.exist()` now.

src/name/resolve.js
```diff
@@ -125,13 +129,32 @@ module.exports = (createCommon, options) => {
...
+
+      return expect(r.substr(0, 6)).to.eql('/ipfs/')
+    })
+
+    it('should fail to resolve /ipns/ipfs.a', async () => {
+      try {
+        await ipfs.name.resolve('ipfs.a')
+      } catch (err) {
+        expect(err).to.exist()
+      }
```
alanshaw: Need to throw here, incase it does resolve!

src/name/resolve.js
hugomrdias: i just prefer to not re-use stuff between tests.

can you explain the second part further? go-ipfs passes this test.

src/name/publish.js
hugomrdias: camel case doesn't work alan, this is used directly as http payload for go-ipfs

src/name/resolve.js
```diff
@@ -125,13 +118,46 @@ module.exports = (createCommon, options) => {
...
+      const r = await ipfs.name.resolve('/ipns/ipfs.io/images/ipfs-logo.svg', { recursive: false })
+      return expect(r).to.match(/\/ipns\/.+\/images\/ipfs-logo.svg$/)
+    })
+
+    it('should resolve /ipns/ipfs.io recursive', async () => {
+      const r = await ipfs.name.resolve('ipfs.io', { recursive: true })
+      return expect(r).to.match(/\/ipfs\/.+$/)
+    })
+
+    it('should resolve /ipns/ipfs.io recursive with remainder', async () => {
```
lidel: @hugomrdias we are missing a test like this one (a recursive resolve of `/ipns/<fqdn>/some/file`), but with a path that is inside of a HAMT-sharded root. 

I believe `/ipns/tr.wikipedia-on-ipfs.org/wiki/Anasayfa.html` can be used for test: it is a safe path that will not disappear and Wikipedia is also an important consumer of sharding feature (`/wiki/` is a sharded directory).

FYSA js-ipfs may not support sharding correctly until https://github.com/ipfs/js-ipfs-http-response/pull/19 is merged first.

src/name/publish.js
alanshaw: It shouldn't be - core takes camel case options, if it needs to be converted to dash-case for the query string then `js-ipfs-http-client` needs to convert it and `js-ipfs` HTTP endpoint needs to convert it back. This is consistent with many of the other endpoints.

hugomrdias: @alanshaw can we defer this to another PR ? ill make an issue to track

alanshaw: Sure, but we should get it resolved before the next release as it's a breaking change to a brand new feature otherwise.

alanshaw:
Is this ready for a re-review @hugomrdias or are there more commits you want to add? 😁

src/name/resolve.js
```diff
@@ -32,106 +24,157 @@ module.exports = (createCommon, options) => {
...
-            expect(res).to.exist()
+    after((done) => common.teardown(done))
 
-            const resolveOptions = {
-              nocache: false,
-              recursive: true
-            }
+    it('should resolve /ipns/ipfs.io', async () => {
+      return expect(await ipfs.name.resolve('/ipns/ipfs.io'))
+        .to.match(/\/ipfs\/.+$/)
```
alanshaw: Can we use `is-ipfs`?

src/ping/ping-pull-stream.js
```diff
@@ -59,7 +59,7 @@ module.exports = (createCommon, options) => {
       )
     })
 
-    it('should fail when pinging an unknown peer over pull stream', (done) => {
+    it.skip('should fail when pinging an unknown peer over pull stream: go-ipfs return success with text: Looking up peer QmUmaEnH1uMmvckMZbh3yShaasvELPW4ZLPWnB4entMTEn', (done) => {
```
alanshaw: Needs to be skipped in `ipfs-http-client` https://github.com/ipfs/js-ipfs-http-client/blob/master/test/interface.spec.js#L237

src/ping/ping-readable-stream.js
```diff
@@ -67,7 +67,7 @@ module.exports = (createCommon, options) => {
       )
     })
 
-    it('should fail when pinging peer that is not available over readable stream', (done) => {
+    it.skip('should fail when pinging peer that is not available over readable stream: go-ipfs returns success', (done) => {
```
alanshaw: Needs to be skipped in `ipfs-http-client` https://github.com/ipfs/js-ipfs-http-client/blob/master/test/interface.spec.js#L237

src/ping/ping.js
```diff
@@ -52,7 +52,7 @@ module.exports = (createCommon, options) => {
       })
     })
 
-    it('should fail when pinging a peer that is not available', (done) => {
+    it.skip('should fail when pinging a peer that is not available: go-ipfs returns success', (done) => {
```
alanshaw: Needs to be skipped in `ipfs-http-client` https://github.com/ipfs/js-ipfs-http-client/blob/master/test/interface.spec.js#L237

src/ping/ping-pull-stream.js
hugomrdias: Can I ask why? If it is skipped here it's skipped everywhere

alanshaw: You want to skip it in the JS IPFS tests too? We need some context for why this is skipped. From the change above it looks like it's skipped because it's not currently working on the latest go-ipfs but if both go-ipfs and js-ipfs are not working then it's fine to skip it here but we just need a comment or something to explain why!

alanshaw: ...and ideally a new issue in JS IPFS to resolve - I'm not sure why this would be failing now - we've not made any changes to ping!

hugomrdias:
Create an issue for the ping problem in js-ipfs 
https://github.com/ipfs/js-ipfs/issues/2186

src/name/resolve.js
hugomrdias: we cant for the remainder tests i would prefer to keep them all the same

PR #460, opened 2019-04-24
add tests for refs endpoint

dirkmc:


src/files-regular/refs-local-tests.js
```diff
@@ -0,0 +1,60 @@
...
+        content('holmes.txt')
+      ]
+
+      ipfs.add(dirs, (err, res) => {
+        expect(err).to.not.exist()
+
+        ipfsRefsLocal(ipfs)((err, refs) => {
+          expect(err).to.not.exist()
+
+          const cids = refs.map(r => r.Ref)
```
alanshaw: Please no PascalCase in core. The HTTP API code translates camelCase to PascalCase and the HTTP API client translates PascalCase back to camelCase.

src/files-regular/refs-local-tests.js
```diff
@@ -0,0 +1,60 @@
...
+          expect(cids.includes('QmVwdDCY4SPGVFnNCiZnX5CtzwWDn6kAM98JXzKxE3kCmn')).to.eql(true)
+          expect(cids.includes('QmR4nFjTu18TyANgC65ArNWp5Yaab1gPzQ4D8zp7Kx3vhr')).to.eql(true)
```
alanshaw: Use `expect(cids).to.include('QmR4nFjTu18TyANgC65ArNWp5Yaab1gPzQ4D8zp7Kx3vhr')` for a better error message when this fails.

e.g.

expected false to be true

vs

expected ['QmVwdDCY4SPGVFnNCiZnX5CtzwWDn6kAM98JXzKxE3kCmn'] to include 'QmR4nFjTu18TyANgC65ArNWp5Yaab1gPzQ4D8zp7Kx3vhr'

src/files-regular/refs-tests.js
```diff
@@ -0,0 +1,315 @@
+/* eslint-env mocha */
+'use strict'
+
+const async = require('async')
```
alanshaw: Now that "async" is a language keyword this can get confusing.

```suggestion
const map = require('async/map')
```

src/files-regular/refs-tests.js
```diff
@@ -0,0 +1,315 @@
...
+    async.map(sorted, ([name, child], cb) => {
+      loadContent(ipfs, child, (err, cid) => {
+        cb(err, { name, cid: cid && cid.toString() })
+      })
+    }, (err, res) => {
+      if (err) {
+        return callback(err)
+      }
+
+      ipfs.object.put({
```
alanshaw: Can we use the DAG API instead?

src/files-regular/refs-tests.js
```diff
@@ -0,0 +1,315 @@
...
+        'atlantic.txt',
+        'indian.txt',
+        'fruits',
+        'tropical.txt',
+        'mushroom.txt'
+      ]
+    },
+
+    'max depth of 1': {
+      params: { format: '<linkname>', r: true, 'max-depth': 1 },
```
alanshaw: `max-depth` => `maxDepth`

src/files-regular/refs-tests.js
```diff
@@ -0,0 +1,315 @@
...
+      expected: [
+        'QmYEJ7qQNZUvBnv4SZ3rEbksagaan3sGvnUq948vSG8Z34',
+        'QmUXzZKa3xhTauLektUiK4GiogHskuz1c57CnnoP4TgYJD',
+        'QmYLvZrFn8KE2bcJ9UFhthScBVbbcXEgkJnnCBeKWYkpuQ',
+        'QmRfqT4uTUgFXhWbfBZm6eZxi2FQ8pqYK5tcWRyTZ7RcgY'
+      ]
+    },
+
+    'prints files in edges format': {
+      params: { e: true },
```
alanshaw: Please no shorthand params in core!

src/files-regular/refs-local-tests.js
dirkmc: hmm I didn't know about that, thanks 👍

src/object/links.js
```diff
@@ -162,5 +162,42 @@ module.exports = (createCommon, options) => {
...
+              anotherLink: { '/': hashes[1] }
+            }
+          }
+          ipfs.dag.put(obj, (err, cid) => {
+            expect(err).to.not.exist()
+            ipfs.object.links(cid, (err, links) => {
+              expect(err).to.not.exist()
+              expect(links.length).to.eql(2)
+
+              // TODO: js-ipfs succeeds but go returns empty strings for link name
```
dirkmc: @alanshaw do you think we should ping the go team about this?

alanshaw: Yes - would you mind opening an issue?

dirkmc: Sure: https://github.com/ipfs/go-ipld-cbor/issues/61

PR #461, opened 2019-05-02
feat: add tests for add data using File DOM api

hugomrdias:
depends on:
https://github.com/hugomrdias/js-ipfs-utils


dirkmc:
The code for these three PRs LGTM 👍 

I wonder if it would make sense to put the shared util code into this repo (interface-js-ipfs-core) instead of having another repo (js-ipfs-utils), so that there is less overhead in terms of releases etc when there are changes made to the utils code?

hugomrdias:
Yeah I also thought of that, and discussed a bit with Alan and because this repo is only tests a new repo seemed appropriate. Release overhead is a strong point I can revert utils into this repo if we all agree.
Please take a look at the ipfs-utils tests and give feedback, I introduced there the concept of self skipping tests, so we don't need to skip multiple times in the interface-core implementations.

package.json
```diff
@@ -46,12 +46,14 @@
     "es6-promisify": "^6.0.1",
     "hat": "0.0.3",
     "into-stream": "^4.0.0",
+    "ipfs": "~0.35.0",
```
alanshaw: Was this intentional?

package.json
```diff
@@ -46,12 +46,14 @@
...
     "ipfs-block": "~0.8.0",
     "ipfs-unixfs": "~0.1.16",
     "ipld-dag-cbor": "~0.13.1",
     "ipld-dag-pb": "~0.15.0",
     "is-ipfs": "~0.5.1",
     "is-plain-object": "^2.0.4",
+    "js-ipfs-utils": "github:hugomrdias/js-ipfs-utils#master",
```
alanshaw: Can the module name be `ipfs-utils` please?

src/files-regular/add.js
```diff
@@ -125,7 +138,7 @@ module.exports = (createCommon, options) => {
 
       ipfs.add(data, (err) => {
         expect(err).to.exist()
-        expect(err.message).to.contain('invalid input')
+        expect(err.message).to.contain('Input not supported')
```
alanshaw: Does this pass when these tests are run against `js-ipfs-http-client`?

alanshaw:
ping @hugomrdias

src/files-regular/add.js
hugomrdias: yes

PR #471, opened 2019-05-15
Fix: use fixtures for refs tests

dirkmc:


src/config/set.js
```diff
@@ -66,7 +66,7 @@ module.exports = (createCommon, options) => {
         expect(err).to.not.exist()
         ipfs.config.get(key, function (err, result) {
           expect(err).to.not.exist()
-          expect(result).to.equal(val)
+          expect(parseInt(result)).to.equal(val)
```
alanshaw: If this is not a number it needs to be fixed in js-ipfs. Please [skip this test](https://github.com/ipfs/interface-js-ipfs-core#skipping-specific-tests) in https://github.com/ipfs/js-ipfs/pull/2004 and open an issue and we'll revisit.

PR #468, opened 2019-05-13
chore: update cids and ipld

alanshaw:

PR #473, opened 2019-05-19
chore: update ipld formats and cid versions

achingbrain:
May need merging with or replacing by #468.


achingbrain:
I didn't realise the `should get links from CBOR object` test had been removed - I wondered how it had ever passed as there was no implementation in `js-ipfs` so, uh, I added it in ipfs/js-ipfs#2068

PR #506, opened 2019-07-25
Added in a test for the opts parameter.

obo20:
The opts parameter wasn't correctly passed in when calling ipfs.object.stat. This is the test that makes sure opts is being correctly passed to through the send function to the IPFS daemon. 

I've also removed the "base58" encoding that was being passed in for some tests as base58 isn't a valid option for encoding and would provide this error: 

Uncaught AssertionError: expected [Error: invalid encoding: base58] to not exist

### DIFF
--- a/SPEC/OBJECT.md
+++ b/SPEC/OBJECT.md
@@ -198,7 +198,8 @@ A great source of [examples][] can be found in the tests for this API.
 
 `options` is a optional argument of type object, that can contain the following properties:
 
-- `enc`, the encoding of multihash (base58, base64, etc), if any.
+- `enc`, The encoding type the output should beencoded with (json, xml, or text), if any.
+- `timeout`, A timeout to pass to the IPFS daemon so the request expires after a certain amount of time.
 
 `callback` must follow `function (err, stats) {}` signature, where `err` is an error if the operation was not successful and `stats` is an Object with following format:
 
@@ -220,7 +221,7 @@ If no `callback` is passed, a [promise][] is returned.
 ```JavaScript
 const multihash = 'QmPTkMuuL6PD8L2SwTwbcs1NPg14U8mRzerB1ZrrBrkSDD'
 
-ipfs.object.stat(multihash, (err, stats) => {
+ipfs.object.stat(multihash, {timeout: '10s'}, (err, stats) => {
   if (err) {
     throw err
   }
--- a/SPEC/OBJECT.md
+++ b/SPEC/OBJECT.md
@@ -198,7 +198,6 @@ A great source of [examples][] can be found in the tests for this API.
 
 `options` is a optional argument of type object, that can contain the following properties:
 
-- `enc`, The encoding type the output should beencoded with (json, xml, or text), if any.
 - `timeout`, A timeout to pass to the IPFS daemon so the request expires after a certain amount of time without any response. NOTE: not yet supported in JS IPFS.
 
 `callback` must follow `function (err, stats) {}` signature, where `err` is an error if the operation was not successful and `stats` is an Object with following format:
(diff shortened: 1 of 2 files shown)


src/object/stat.js
```diff
@@ -80,6 +80,30 @@ module.exports = (createCommon, options) => {
 
       expect(expected).to.deep.equal(stats)
     })
+    
+    it('should pass on the opts', (done) => {
```
alanshaw: We're specifically testing that `timeout` is passed:

```suggestion
    it('should respect timeout option', (done) => {
```

src/object/stat.js
```diff
@@ -80,6 +80,30 @@ module.exports = (createCommon, options) => {
...
+      ipfs.object.put(testObj, (err, cid) => {
+        expect(err).to.not.exist()
+        const timeout = 2;
+        const startTime = new Date();
+        const badCid = 'QmNggDXca24S6cMPEYHZjeuc4QRmofkRrAEqVL3MzzzzzZ';
+
+        //we can test that we are passing in opts by testing the timeout option for a CID that doesn't exist
+        ipfs.object.stat(badCid, {timeout: `${timeout}s`}, (err, stats) => {
+          let timeForRequest = (new Date () - startTime)/1000;
+          console.log(err);
```
alanshaw:
```suggestion
          expect(err).to.exist()
```

SPEC/OBJECT.md
```diff
@@ -198,7 +198,8 @@ A great source of [examples][] can be found in the tests for this API.
 
 `options` is a optional argument of type object, that can contain the following properties:
 
-- `enc`, the encoding of multihash (base58, base64, etc), if any.
+- `enc`, The encoding type the output should beencoded with (json, xml, or text), if any.
```
alanshaw: I think this should just be removed - it's not documented on docs.ipfs.io and the js-ipfs http and cli do no respect it. I think it's a hangover from pre-CID days.

It's not the same as the `encoding` flag: https://docs.ipfs.io/reference/api/http/#flags

SPEC/OBJECT.md
```diff
@@ -198,7 +198,8 @@ A great source of [examples][] can be found in the tests for this API.
...
+- `timeout`, A timeout to pass to the IPFS daemon so the request expires after a certain amount of time.
```
alanshaw:
```suggestion
- `timeout`, A timeout to pass to the IPFS daemon so the request expires after a certain amount of time without any response. NOTE: not yet supported in JS IPFS.
```

src/object/stat.js
```diff
@@ -81,6 +81,30 @@ module.exports = (createCommon, options) => {
...
+        const startTime = new Date()
+        const badCid = 'QmNggDXca24S6cMPEYHZjeuc4QRmofkRrAEqVL3MzzzzzZ'
+
+        // we can test that we are passing in opts by testing the timeout option for a CID that doesn't exist
+        ipfs.object.stat(badCid, { timeout: `${timeout}s` }, (err, stats) => {
+          const timeForRequest = (new Date() - startTime) / 1000
+          expect(err).to.exist()
+          expect(err.message).to.equal('failed to get block for QmNggDXca24S6cMPEYHZjeuc4QRmofkRrAEqVL3MzzzzzZ: context deadline exceeded')
+          expect(stats).to.not.exist()
+          expect(timeForRequest).to.not.lessThan(timeout)
```
alanshaw: Should we also have an upper bound? Not more than {timeout + 1}s?

obo20: Done

obo20:
@alanshaw all linting changes have been made and everything you've commented on should be resolved.

PR #514, opened 2019-08-27
fix: tests of addFromURL in browsers

lidel:
> Required by https://github.com/ipfs/js-ipfs/pull/2304

Old version of `addFromURL` tests was Node-centric and did not work in web browser contexts.

This PR:

- moves the HTTP server to a separate file to enable spawning it from
  aegir hook (see .aegir.js/hooks/browser/pre|post in js-ipfs from https://github.com/ipfs/js-ipfs/pull/2304)
- removes  overhead of starting and stopping HTTP servers multiple times
- shortens/simplifies test code a bit

### DIFF
--- a/src/utils/echo-http-server.js
+++ b/src/utils/echo-http-server.js
@@ -4,7 +4,7 @@
 const http = require('http')
 const https = require('https')
 const URL = require('url').URL || self.URL
-const { isBrowser, isWebWorker } = require('ipfs-utils/src/env')
+const { isNode } = require('ipfs-utils/src/env')
 
 const loadFixture = require('aegir/fixtures')
 const sslOpts = Object.freeze({
@@ -14,8 +14,6 @@ const sslOpts = Object.freeze({
 
 const httpPort = 11080
 const httpsPort = 11443
-module.exports.httpPort = httpPort
-module.exports.httpsPort = httpsPort
 
 // Create a mock of remote HTTP server that can return arbitrary text in response
 // or redirect to other URL. Used in tests of ipfs.addFromURL etc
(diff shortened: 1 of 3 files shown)


src/utils/echo-http-server.js
```diff
@@ -0,0 +1,86 @@
...
+  key: loadFixture('test/fixtures/ssl/privkey.pem', 'interface-ipfs-core'),
+  cert: loadFixture('test/fixtures/ssl/cert.pem', 'interface-ipfs-core')
+})
+
+const httpPort = 11080
+const httpsPort = 11443
+
+// Create a mock of remote HTTP server that can return arbitrary text in response
+// or redirect to other URL. Used in tests of ipfs.addFromURL etc
+module.exports.createServer = (opts) => {
```
hugomrdias: can this `https://github.com/lukechilds/create-test-server` be helpful here ?

lidel: not much, most of the code is unusual and specific to our logic and would not benefit from the lib

alanshaw:
@lidel it took me a while to figure out that there's a noop server in the browser. What do you think about requiring the echo server to be started in all environments? 1) It would simplify the setup and teardown code in the tests here, 2) we wouldn't need a noop server (potentially helping with any misunderstandings) and 3) it would only need a couple of changes to [.aegir.js](https://github.com/ipfs/js-ipfs/pull/2304/files#diff-a956444d0c75368b8d347072262fb445) (which you're already editing in that PR) to also start it for Node.js.

lidel:
@alanshaw if we are going to simplify this, I'd like to remove HTTPS echo server and have only HTTP instance, as suggested in https://github.com/ipfs/js-ipfs/pull/2304#issuecomment-525777230. I fail to see what is the value of testing HTTPS.

Y/n?

alanshaw:
@lidel go for it!

lidel:
@alanshaw check  https://github.com/ipfs/interface-js-ipfs-core/pull/514/commits/a810adfcbb977257851a4be3aae158a350119810 and https://github.com/ipfs/js-ipfs/pull/2304/commits/32e76cc1d05e88d5233293d4dc6dc225b9ed2d33

lidel:
(rebased on master to include pubsub fixes)

PR #516, opened 2019-08-30
chore: skip should publish message from string

hugomrdias:


alanshaw:
ah, I meant skip from js-ipfs - there’s no need to skip in ipfs-http-client as it passes here!

alanshaw:
https://github.com/ipfs/interface-js-ipfs-core#skipping-specific-tests

PR #523, opened 2019-09-06
fix(ci): avoid fs clash during ipfs.add* on Windows

lidel:
This PR fixes random TravisCI failures on Windows and unblocks https://github.com/ipfs/js-ipfs/pull/2304

### Root cause

Windows issue was triggered by parallel `ipfs.add*` of the same data:

```js
      parallel({
        result: (cb) => ipfs.addFromURL(url, cb),
        expectedResult: (cb) => ipfs.add(Buffer.from(text), cb)
      }, (err, { result, expectedResult }) => {
        expect(err).to.not.exist()
```

Both `addFromURL` and `add` added the same data, touched same files on filesystem, triggering a known problem with Node on Windows and Travis:

- https://travis-ci.community/t/eperm-operation-not-permitted-nodejs-electron/2571

### Fix 

One way to fix this type of error is to run as Windows Administrator. It does not seem to be feasible for TravisCI.

Instead, switching to sequential mode of operation removes `addFromURL` tests as a surface for file system race conditions on Windows.

### DIFF
--- a/src/files-regular/add-from-url.js
+++ b/src/files-regular/add-from-url.js
@@ -2,7 +2,7 @@
 'use strict'
 
 const { getDescribe, getIt, expect } = require('../utils/mocha')
-const parallel = require('async/parallel')
+const series = require('async/series')
 const { echoUrl, redirectUrl } = require('../utils/echo-http-server')
 
 module.exports = (createCommon, options) => {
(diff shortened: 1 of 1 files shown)


alanshaw:
The fix for this is https://github.com/ipfs/js-datastore-fs/pull/27#event-2616223483

hugomrdias:
> The fix for this is [ipfs/js-datastore-fs#27](https://github.com/ipfs/js-datastore-fs/pull/27)

That was merged into 0.8.x branch and ipfs-repo is on 0.9.x 
@achingbrain do we need to port it to master ?

achingbrain:
The PR for the 0.9.x branch is https://github.com/ipfs/js-datastore-fs/pull/28

lidel:
Should we close this, and bubble up the datastore fix instead?

lidel:
Closing, as https://github.com/ipfs/js-ipfs/pull/2304 confirmed this is fixed in js-datastore-fs

PR #540, opened 2019-10-08
feat: support for reconstructed CIDs of refs-local

AuHau:
Related to https://github.com/ipfs/js-ipfs/issues/2415


SPEC/REPO.md
```diff
@@ -20,7 +20,7 @@ Where:
 - `err` is an Error if the whole GC operation was not successful.
 - `res` is an array of objects that contains the following properties
   - `err` is an Error if it was not possible to GC a particular block.
-  - `cid` is the [CID][cid] of the block that was Garbage Collected.
+  - `multihash` is the [multihashes][multihashes] of the block that was Garbage Collected.
```
achingbrain: If the text in the square brackets is the same as the shortcut in the footer, I think this can just be `[multihashes][]`

AuHau: Oh did not know that! That is neat! Thanks for the heads up!

PR #550, opened 2019-10-31
docs: clarify MFS cp/mv from IPFS path

alanshaw:
Copying or moving from an IPFS path does _not_ recursively retrieve all contents from the network if not in your node's repo - it only retrieves the root node.


achingbrain:
Maybe resolving the entire graph on `cp`/`mv` is something we could add as an option to `ipfsx`?

PR #552, opened 2019-11-05
RFC: unite the files APIs

alanshaw:
**Not finished yet!**

### DIFF
--- a/SPEC/FILESv2.md
+++ b/SPEC/FILESv2.md
@@ -1,0 +1,55 @@
+# Files API v2
+
+There have long been dreams of uniting the files APIs. This is a new proposal for doing so.
+
+## Background
+
+In IPFS there are two sets of APIs for dealing with files. At the root level there is `add`, `cat`, `get` and `ls`. Since IPFS deals with immutable data there's no APIs to change data once it has been imported. You can only retrieve data, import more data or remove data you no longer wish to store in your local repo (That last part refers to pinning and garbage collection, which is somewhat outside the scope of this proposal but elements of these things will be covered with respect to UX and performance. If you're unfamiliar with those concepts, please read https://docs.ipfs.io/guides/concepts/pinning/ before continuing).
+
+You _could_ change a file outside of IPFS and re-import it. IPFS will do it's best to de-dupe the data, but the fact remains that you haven't changed the original file. Changing files is quite a common thing to do so there exists a separate set of API methods specifically for dealing with changing files that have already been imported into IPFS. These are the Mutable File System (or MFS) API methods that deal with the fiddly work of mutating a DAG to change it's structure while re-using as many existing nodes as possible. It is an abstraction to make the immutable nature of IPFS appear as though it is mutable, but behind the scenes the only mutation that occurs is the tracking of the CID for a root node of a DAG that contains all MFS data in _your_ IPFS repo. Nodes in the DAG are never changed, only created or removed as necessary.
+
+The MFS API methods are `write`, `read`, `ls`, `cp`, `mv`, `mkdir`, `rm`, `mkdir`, `stat`, `flush`. These API methods reside under the `files` namespace.
+
+### IPFS paths and MFS paths
+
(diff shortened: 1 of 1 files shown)


SPEC/FILESv2.md
```diff
@@ -0,0 +1,237 @@
...
+| `ipfs import` | ❌ | ❌ |
+| `ipfs ls` | ✅ | ✅ |
+| `ipfs mkdir` | ❌ | ✅ |
+| `ipfs mv` | ✅ | ✅ |
+| `ipfs read` | ✅ | ✅ |
+| `ipfs rm` | ❌ | ✅ |
+| `ipfs stat` | ✅ | ✅ |
+| `ipfs write` | ❌ | ✅ |
+
+The `/ipfs` directory in MFS problem can simply be avoided by either assuming IPFS path (the current solution) or by denying writes to a directory of this name.
```
Stebalien: If we're going to join `mfs` with everything else, I'd like to namespace MFS under something like `/local` or `/files. That way, we can continue adding new namespaces without clobbering user-defined directories.

* /ipfs
* /ipns
* /ipld?
* /p2p?? (might be nice to expose this as sockets when mounted with fuse).
* /git/Cid/... ???

SPEC/FILESv2.md
```diff
@@ -0,0 +1,237 @@
...
+
+### 6. Streaming APIs by default
+
+The file system APIs will be streaming by default. Due to the way we store and retrieve data it makes sense for our API methods to stream content when retrieving it locally or over the network. Buffering APIs can cause OOM issues, give no feedback to the user on progress and they can be trivially wrapped to collect all items in order to achieve the same effect as a buffering API.
+
+Streaming APIs will use a language native / standard library feature that is supported in all runtimes that IPFS is actively targeting. This prevents bloat and by only supporting one streaming mechanism it reduces API surface area.
```
Stebalien: Note: Let's also stop streaming _bytes_. If we switch to only streaming _objects_, we can stream metadata (progress, etc.).

daviddias:
YES! ❤️ Will review more carefully soon. Prev work for context to others at https://github.com/ipfs/specs/issues/98 && https://github.com/ipfs/interface-js-ipfs-core/issues/284

SPEC/FILESv2.md
```diff
@@ -0,0 +1,237 @@
...
+| `ipfs ls` | (removed) |
+| `ipfs files cp` | `ipfs cp` |
+| `ipfs files flush` | `ipfs flush` |
+| `ipfs files ls` | `ipfs ls` |
+| `ipfs files mkdir` | `ipfs mkdir` |
+| `ipfs files mv` | `ipfs mv` |
+| `ipfs files read` | `ipfs read` |
+| `ipfs files rm` | `ipfs rm` |
+| `ipfs files stat` | `ipfs stat` |
+| `ipfs files write` | `ipfs write` |
```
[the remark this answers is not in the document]
achingbrain: The js `files.write` implementation takes a `strategy` parameter that can be `balanced` or `trickle` (because `ipfs.add` and `ipfs.files.write` both use the same `unixfs-importer` which receives the parameter).

If you `files.write` a file, it'll get converted into a `trickle` DAG, unless you pass `strategy=balanced` in which case it'll convert it into a `balanced` DAG (rebalancing as necessary).

I think we could standardise on one and allow the user to override as they see fit.

If we could then extend the [UnixFS v1.5 metadata proposal](https://github.com/ipfs/specs/pull/223) to include the originally selected DAG strategy we'd be able to optimise for appending to trickle DAGs.

SPEC/FILESv2.md
```diff
@@ -0,0 +1,237 @@
...
+# ...
+```
+
+It's also the [current name of the API method to import data into `go-filecoin`](https://github.com/filecoin-project/go-filecoin/blob/204837f72d20bd89889fcf92061c8846b238ccf4/cmd/go-filecoin/client.go#L62), which is reassuring.
+
+#### 1.1 Always add to MFS
+
+Adding files to MFS removes any performance overhead of creating/maintaining pinset DAG nodes, unburdens the user from understanding pinning (for the most part), improves visibility of added files and makes it significantly easier to find or remove files that were previously imported.
+
+The `ipfs import` command _optionally_ takes a MFS path option `--dest`, a directory into which imported files are placed. Note the destination directory is automatically created (but not any parents). If the destination directory exists already then an error is thrown, unless the `--overwrite` flag is set. This causes any existing files with the same name as the imported files to be overwritten.
```
achingbrain: Small point, but maybe `--force` instead of `--overwrite`?  Feels a bit more unixy.

PR #553, opened 2019-11-05
feat: test ipns resolve of peerid as cid

lidel:
This PR adds IPFS test for resolving paths with libp2p-key represented as CIDv1 Base32.

Context:
- https://github.com/libp2p/specs/blob/master/RFC/0001-text-peerid-cid.md
- https://github.com/ipfs/js-ipfs/pull/2566

cc https://github.com/libp2p/specs/issues/216, https://github.com/ipfs/ipfs/issues/337

### DIFF
--- a/src/name/resolve.js
+++ b/src/name/resolve.js
@@ -5,6 +5,7 @@
 const { spawnNodeWithId } = require('../utils/spawn')
 const { getDescribe, getIt, expect } = require('../utils/mocha')
 const delay = require('../utils/delay')
+const CID = require('cids')
 
 module.exports = (createCommon, options) => {
   const describe = getDescribe(options)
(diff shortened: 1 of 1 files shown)


src/name/resolve.js
```diff
@@ -45,6 +46,20 @@ module.exports = (createCommon, options) => {
         .to.eq(`/ipfs/${path}`)
     })
 
+    it('should resolve a record from peerid as cidv1 in base32', async function () {
+      this.timeout(20 * 1000)
+      const [{ path }] = await ipfs.add(Buffer.from('should resolve a record from cidv1b32'))
```
alanshaw: Would it be better to use `hash` here?

alanshaw: ...all the others are using `path`, lets leave it for another day!

PR #554, opened 2019-11-08
test: add human readable repo stats test

PedroMiguelSS:
Add human-readable repo stats test.

### DIFF
--- a/src/repo/stat.js
+++ b/src/repo/stat.js
@@ -1,7 +1,7 @@
 /* eslint-env mocha */
 'use strict'
 
-const { expectIsRepo } = require('../stats/utils')
+const { expectIsRepo, expectIsRepoHumanized } = require('../stats/utils')
 const { getDescribe, getIt, expect } = require('../utils/mocha')
 
 module.exports = (createCommon, options) => {
@@ -40,6 +40,12 @@ module.exports = (createCommon, options) => {
       return ipfs.repo.stat().then((res) => {
         expectIsRepo(null, res)
       })
+    })
+
+    it('should get human readable repo stats', async () => {
+      const res = await ipfs.repo.stat({ human: true })
+
+      expectIsRepoHumanized(null, res)
     })
   })
 }
--- a/src/repo/stat.js
+++ b/src/repo/stat.js
@@ -1,7 +1,7 @@
 /* eslint-env mocha */
 'use strict'
 
-const { expectIsRepo, expectIsRepoHumanized } = require('../stats/utils')
+const { expectIsRepo } = require('../stats/utils')
 const { getDescribe, getIt, expect } = require('../utils/mocha')
 
 module.exports = (createCommon, options) => {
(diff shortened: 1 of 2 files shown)


src/repo/stat.js
```diff
@@ -41,5 +41,11 @@ module.exports = (createCommon, options) => {
         expectIsRepo(null, res)
       })
     })
+
+    it('should get human readable repo stats', async () => {
+      const res = await ipfs.repo.stat({ human: true })
+
+      expectIsRepoHumanized(null, res)
```
alanshaw: Please move these assertions into this test. The function is not being reused so there's no need for it to be extracted.

src/stats/utils.js
```diff
@@ -57,3 +57,18 @@ exports.expectIsRepo = (err, res) => {
...
+exports.expectIsRepoHumanized = (err, res) => {
+  expect(err).to.not.exist()
+  expect(res).to.exist()
+  expect(res).to.have.a.property('numObjects')
+  expect(res).to.have.a.property('repoSize')
+  expect(res).to.have.a.property('repoPath')
+  expect(res).to.have.a.property('version')
+  expect(res).to.have.a.property('storageMax')
+  expect(res.numObjects).to.be.a('number')
+  expect(res.repoSize).to.eql('30.1 kB')
```
alanshaw: This is a brittle assertion - it may work for now, but it could easily change in the future (the default assets in a repo could be updated for example) causing this test to fail.

We should instead assert that this value is in the format expected since an actual value is almost impossible to test for without mocking.

PedroMiguelSS: Got it. I'll use a regex to test the expected format.

PedroMiguelSS:
Can I get a second review here, please? @alanshaw @hugomrdias

src/repo/stat.js
```diff
@@ -41,5 +41,21 @@ module.exports = (createCommon, options) => {
...
+
+      expect(res).to.exist()
+      expect(res).to.have.a.property('numObjects')
+      expect(res).to.have.a.property('repoSize')
+      expect(res).to.have.a.property('repoPath')
+      expect(res).to.have.a.property('version')
+      expect(res).to.have.a.property('storageMax')
+      expect(res.numObjects).to.be.a('number')
+      expect(res.repoSize).to.match(/\s.?B$/gm)
+      expect(res.storageMax).to.match(/\s.?B$/gm)
```
alanshaw: `/\s.?B$/gm` - hmm that doesn't look right to me.

Screenshot 2019-11-18 at 15 32 19

Are the `gm` modifiers needed? Does the test pass? If not, please ensure the test passes locally before reuesting a review. If you need any help linking and running only this test then ping me on IRC.

PedroMiguelSS: Yap, sorry, the modifiers are not needed. Furthermore, I'll update the expression, I'll make it more specific.
I wouldn't have asked for a second review if the test was not passing locally. You can see it passing here:
image


This is the `res` object

```js
{
   ...
   repoSize: '30.1 KB'
   storageMax: '9.01 PB'
}
```

My regex was matching: 
image
and
image

What do you think about this regex? Does this one seem better? It would match the whole string instead of just the units.
`/^[\d]+(\.[\d]+)?\s[KMGTP]?B$/`

Considering the `res` object (mentioned above), this new regex will match:
image
and
image

alanshaw:
This can be closed now right @PedroMiguelSS?

PedroMiguelSS:
Closed in favor of https://github.com/ipfs/js-ipfs/pull/2630.

PR #562, opened 2019-11-20
chore: async await refactor

### LINKED ISSUE #526: Convert tests to use `async/await`
As part of the [great migration to async iterators](https://github.com/ipfs/js-ipfs/issues/1670) we need to convert our tests to use async/await.  This is often best done separately from converting the tested code from callbacks to promises.

This repo's tests need converting.

PedroMiguelSS:
This PR converts the tests to use async/await syntax.

Closes #526.

The tests' setup was not changed because https://github.com/ipfs/interface-js-ipfs-core/pull/541 is already doing that. 

This PR replaces the old https://github.com/ipfs/interface-js-ipfs-core/pull/547. The difference is that this one doesn't depend on [`feat/interface-tests`](https://github.com/ipfs/interface-js-ipfs-core/tree/feat/interface-tests) branch, which depends on a currently WIP branch - ipfs/js-ipfsd-ctl#395. 

This way we can merge this refactor as soon as it gets approved.


alanshaw:
Thank @PedroMiguelSS - just a minor point for the future, these commits should be prefixed with `refactor:` not `chore:`. https://github.com/ipfs/community/blob/master/CONTRIBUTING_JS.md#type has all the possible prefixes and when to use them.

I'll take a look at this now.

PR #564, opened 2019-11-28
chore: add tests timeouts as they were previously

PedroMiguelSS:
This PR adds `timeouts` as they were before the async/await refactor.


src/bitswap/wantlist.js
```diff
@@ -5,42 +5,58 @@
...
 
-    before(async () => {
+    before(async function () {
+      // CI takes longer to instantiate the daemon, so we need to increase the
+      // timeout for the before step
+      this.timeout(60 * 1000)
+
       ipfsA = await common.setup()
-      ipfsB = await common.setup()
+      ipfsB = await common.setup({ factoryOptions: { type: 'go' } })
```
achingbrain: Why is this specifying go?

src/bitswap/wantlist.js
```diff
@@ -5,42 +5,58 @@
...
-    it('should get the wantlist', () => {
-      return waitForWantlistKey(ipfsB, key)
+      return common.teardown()
     })
 
-    it('should get the wantlist by peer ID for a diffreent node', () => {
-      return waitForWantlistKey(ipfsA, key, { peerId: ipfsB.peerId.id })
+    it('should get the wantlist', function () {
+      this.timeout(TIMEOUT)
+      return waitForWantlistKey(ipfsB, key, { timeout: TIMEOUT })
```
achingbrain: Do we need to pass the timeout here?  If the request doesn't finish in time the test will fail anyway and it's not supported by core, only the rpc client.

src/bitswap/wantlist.js
```diff
@@ -5,42 +5,58 @@
...
     })
 
-    it('should not get the wantlist when offline', async () => {
+    it('should get the wantlist by peer ID for a different node', function () {
+      this.timeout(TIMEOUT)
+      return waitForWantlistKey(ipfsA, key, {
+        peerId: ipfsB.peerId.id,
+        timeout: TIMEOUT
```
achingbrain: Do we need to pass the timeout here?  If the request doesn't finish in time the test will fail anyway and it's not supported by core, only the rpc client.

src/bitswap/wantlist.js
PedroMiguelSS: This was a change trying to fix a failing test (only on CI, the test is passing locally). But I'll remove it because it didn't solve the problem.

PR #565, opened 2019-12-02
refactor: http-server util file

PedroMiguelSS:
This PR refactors echo server `start()` and `stop()` from `callback` to `promise`.

### DIFF
--- a/src/utils/echo-http-server.js
+++ b/src/utils/echo-http-server.js
@@ -43,15 +43,11 @@ module.exports.createServer = () => {
 
   const server = http.createServer(handler)
 
-  server.start = (opts, cb) => {
-    if (typeof opts === 'function') {
-      cb = opts
-      opts = {}
-    }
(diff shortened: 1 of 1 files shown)


src/utils/echo-http-server.js
```diff
@@ -43,15 +43,11 @@ module.exports.createServer = () => {
...
   const server = http.createServer(handler)
 
-  server.start = (opts, cb) => {
-    if (typeof opts === 'function') {
-      cb = opts
-      opts = {}
-    }
-    return server.listen(Object.assign({ port: defaultPort, host: '127.0.0.1' }, opts), cb)
-  }
+  server.start = (opts) => new Promise(
```
achingbrain: This should handle errors while starting up:

```javascript
server.start = (opts) => new Promise((resolve, reject) => {
  server.once('error', reject)
  server.listen(Object.assign({ port: defaultPort, host: '127.0.0.1' }, opts), () => {
    server.removeListener('error', reject)
    resolve()
  })
})
```

Same thing for shutting down.

PR #567, opened 2019-12-03
refactor: async iterables

alanshaw:
One day `it-all` will be https://tc39.es/proposal-iterator-helpers/#sec-asynciteratorprototype-toarray 🤞

See the breaking changes messages in https://github.com/ipfs/js-ipfs-http-client/pull/1183 for a high level overview of what is being changed.

TODO:

- [x] Move `addFromFs` and `addFromURL` into `add` tests using `globSource` and `urlSource` from `ipfs-utils`
- [x] Update docs

resolves https://github.com/ipfs/interface-js-ipfs-core/issues/394


SPEC/DHT.md
```diff
@@ -7,79 +7,80 @@
...
-
 #### `dht.findPeer`
 
-> Retrieve the Peer Info of a reachable node in the network.
+> Find the multiaddresses associated with a Peer ID
 
 ##### `ipfs.dht.findPeer(peerId)`
 
-Where `peerId` is a IPFS/libp2p Id from [PeerId](https://github.com/libp2p/js-peer-id) type.
+Where `peerId` is a Peer ID in `String`, [`CID`](https://github.com/multiformats/js-cid) or [`PeerId`](https://github.com/libp2p/js-peer-id) format.
```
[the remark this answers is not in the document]
alanshaw: > URLSearchParams calls toString on object when serializing ?

Unfortunately it does not - that's why all values that aren't strings or numbers are wrapped in `` `${value}` `` - to get `value.toString()`.

> Some suggestions for the param `peerId`
> - instances of Cid or Peerid 
> - the cidv1 output of `toString` from cid or peerid
> - all of the above and do something like

These are the things the code supports right now, and additonally a CID as a buffer.

SPEC/DHT.md
```diff
@@ -7,79 +7,80 @@
...
 
 #### `dht.get`
 
-> Retrieve a value from DHT
+> Given a key, query the routing system for its best value.
 
 ##### `ipfs.dht.get(key)`
 
-Where `key` is a Buffer.
+Where `key` is a `Buffer`.
```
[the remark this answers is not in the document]
alanshaw: Will fix

SPEC/DHT.md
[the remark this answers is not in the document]
hugomrdias: didn't notice the `` there

SPEC/OBJECT.md
```diff
@@ -129,14 +126,14 @@ A great source of [examples][] can be found in the tests for this API.
 
 | Type | Description |
 | -------- | -------- |
-| `Promise<Buffer>` | A Buffer with the data that the MerkleDAG node contained |
+| `Promise<Buffer>` | An Promise that resolves to Buffer objects with the data that the MerkleDAG node contained |
```
lidel:
```suggestion
| `Promise<Buffer>` | A Promise that resolves to Buffer objects with the data that the MerkleDAG node contained |
```

SPEC/PIN.md
```diff
@@ -53,28 +50,24 @@ A great source of [examples][] can be found in the tests for this API.
...
+  - `type` - filter by this type of pin ("recursive", "direct" or "indirect")
 
 **Returns**
 
 | Type | Description |
 | -------- | -------- |
-| `Promise<Array>` | An array of current pinned objects |
-
-an array of objects with keys `hash` and `type` is returned.
+| `AsyncIterable<{ hash: string, type: string }>` | An async iterable that yields currently pinned objects with `hash` and `type` properties. `hash` is a string CID of the pinned node, `type` is the pin type ("recursive", "direct" or "indirect") |
```
lidel: I think `hash` should be replaced with `cid`, to match https://github.com/ipfs/js-ipfs-http-client/pull/1183:

> BREAKING CHANGE: pin.ls results now contain a cid property (a CID instance) instead of a string hash property.

src/cat.js
```diff
@@ -25,103 +26,103 @@ module.exports = (common, options) => {
...
 
       const cidv0 = new CID(res[0].hash)
       expect(cidv0.version).to.equal(0)
 
       const cidv1 = cidv0.toV1()
 
-      const output = await ipfs.cat(cidv1)
-      expect(output).to.eql(input)
+      const output = await concat(ipfs.cat(cidv1))
+      expect(output.slice()).to.eql(input)
```
lidel: Why we need a shallow copy here (`.slice()`)?

alanshaw: concat returns a BufferList

SPEC/MISCELLANEOUS.md
```diff
@@ -5,13 +5,8 @@
...
-* [pingPullStream](#pingpullstream)
-* [pingReadableStream](#pingreadablestream)
 * [resolve](#resolve)
 
-### ⚠️ Note
-Although not listed in the documentation, all the following APIs that actually return a **promise** can also accept a **final callback** parameter.
-
 #### `id`
 
 > Returns the identity of the Peer
```
achingbrain: Suggestion:  `id` should be a rich object:

```markdown
The returned object contains the following keys:

- `id` (CID) - A v1 [CID][cid] with the codec 'libp2p-key'
- `publicKey` (Buffer) - Bytes that make up the node's public key
- `addresses` (Array) - List of [Multiaddr][multiaddr] instances 
- `agentVersion` (String) - An implementation type identifier
- `protocolVersion` (String) - An implementation version identifier
```

PR #569, opened 2019-12-09
fix: handle err on both start and stop echo-server

PedroMiguelSS:
`echo-server` was not handling properly error on both `start()` and `stop()` methods. 

@achingbrain suggested on https://github.com/ipfs/interface-js-ipfs-core/pull/565 's [review](https://github.com/ipfs/interface-js-ipfs-core/pull/565/files#r352515503):

```javascript
server.start = (opts) => new Promise((resolve, reject) => {
  server.once('error', reject)
  server.listen(Object.assign({ port: defaultPort, host: '127.0.0.1' }, opts), () => {
    server.removeListener('error', reject)
    resolve()
  })
})
```

I think we just want to check `server.listen()` and `server.close()` related errors. For this reason, my suggestion is:

```javascript
server.start = (opts) => new Promise(
  (resolve, reject) => server.listen(
    Object.assign({ port: defaultPort, host: '127.0.0.1' }, opts), (err) => err ? reject(err) : resolve()
  )
)
```

### DIFF
--- a/src/utils/echo-http-server.js
+++ b/src/utils/echo-http-server.js
@@ -44,10 +44,12 @@ module.exports.createServer = () => {
   const server = http.createServer(handler)
 
   server.start = (opts) => new Promise(
-    (resolve) => server.listen(Object.assign({ port: defaultPort, host: '127.0.0.1' }, opts), resolve)
+    (resolve, reject) => server.listen(
+      Object.assign({ port: defaultPort, host: '127.0.0.1' }, opts), (err) => err ? reject(err) : resolve()
+    )
   )
 
-  server.stop = () => new Promise((resolve) => server.close(resolve))
+  server.stop = () => new Promise((resolve, reject) => server.close((err) => err ? reject(err) : resolve()))
 
   return server
 }
--- a/src/utils/echo-http-server.js
+++ b/src/utils/echo-http-server.js
@@ -53,9 +53,9 @@ module.exports.createServer = () => {
 
   server.stop = () => new Promise((resolve, reject) => {
     server.once('error', reject)
-    server.close(() => {
+    server.close((err) => {
       server.removeListener('error', reject)
-      resolve()
+      err ? reject(err) : resolve()
     })
   })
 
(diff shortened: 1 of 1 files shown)


achingbrain:
You need to register for the `error` event as well.  The callback passed to `.listen()` is only added as a listener for the `listening ` event.

https://nodejs.org/api/net.html#net_server_listen

hugomrdias:
promisifing events is tricky business, do you have any example of a use-case where the error coming from `.listen()` isnt enough ?

achingbrain:
My point is there will be no error passed to the callback passed to `.listen()`.

When you start a server you'll either get a `listening` event (happy path) or an `error` event (unhappy path).

hugomrdias:
humm right i did got that the callback is just a listener to the `listening` event. 

This makes more sense now :) plus as we are only wrapping it in a `new Promise` should be safe.
```js
server.start = (opts) => new Promise((resolve, reject) => {
  server.once('error', reject)
  server.listen(Object.assign({ port: defaultPort, host: '127.0.0.1' }, opts), () => {
    server.removeListener('error', reject)
    resolve()
  })
})
```

PedroMiguelSS:
@achingbrain , can you give me a second review here, please?

src/utils/echo-http-server.js
```diff
@@ -43,11 +43,21 @@ module.exports.createServer = () => {
...
+    server.listen(Object.assign({ port: defaultPort, host: '127.0.0.1' }, opts), () => {
+      server.removeListener('error', reject)
+      resolve()
+    })
+  })
 
-  server.stop = () => new Promise((resolve) => server.close(resolve))
+  server.stop = () => new Promise((resolve, reject) => {
+    server.once('error', reject)
+    server.close(() => {
```
hugomrdias: the callback can be called with an Error if the server was not open 

https://github.com/nodejs/node/blob/9cfb3841785cd658bb13369ed642247e089a5533/lib/net.js#L1519

PedroMiguelSS: I saw that on docs but I thought `server.once('error', reject)` would call its callback on any error (including the `[ERR_SERVER_NOT_RUNNING]`).

PedroMiguelSS: Btw, do we really need `error` event in this case?

PR #571, opened 2019-12-12
fix: fix pubsub subscribe with multiple node in firefox

hugomrdias:
In firefox one empty msg is send first to start the flow, this PR accounts for that.

https://github.com/ipfs/js-ipfs-http-client/blob/master/src/pubsub/subscribe.js#L27-L37

### DIFF
--- a/src/pubsub/subscribe.js
+++ b/src/pubsub/subscribe.js
@@ -17,7 +17,7 @@ module.exports = (common, options) => {
   const it = getIt(options)
 
   describe('.pubsub.subscribe', function () {
-    this.timeout(80 * 1000)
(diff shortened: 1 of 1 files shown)


alanshaw:
I was thinking we'd change the empty message to be a special value that the client would recognise and drop silently...this is ugly and I hate it but keeps the hack leaking into other places (like here).

hugomrdias:
close and fix in the clients ?

alanshaw:
...I'm just not 100% sure if it's possible (I didn't think about it too much) - for instance js-ipfs will still receive the message - do we have tests that send from firefox to a in process JS node?

PR #572, opened 2019-12-12
docs: add mode and mtime and .touch/.chmod mfs commands

achingbrain:
Needs more tests adding before this is ready.  Early feedback is encouraged.


SPEC/FILES.md
```diff
@@ -759,6 +789,48 @@ A great source of [examples][] can be found in the tests for this API.
...
+  - An existing MFS path to a file or a directory (e.g. `/my-dir/my-file.txt`)
+  - An IPFS path (e.g. `/ipfs/QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks`)
+  - A [CID][cid] instance (e.g. `new CID('QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks')`)
+- `mode` is the new file mode.  It might be:
+  - A string octal, e.g. `'0755'`
+  - A string modification of the existing mode, e.g. `'+x'`, `'-gw'`, etc
+  - An integer, e.g. the returned value from `parseInt('0755', 8)`
+- `options` is an optional Object that might contain the following keys:
+  - `recursive` is a Boolean value that indicates if `mode` should be applied to all sub files/directories of `path` (default: false)
+  - `format` is what type of nodes to write any modified entries as (default: `dag-pb`)
```
alanshaw: Is this actually possible? What formats are supported? I’ve seen this in the docs for the other commands but always wondered...

SPEC/FILES.md
```diff
@@ -58,6 +60,8 @@ Where `data` may be:
 {
     path: '/tmp/myfile.txt', // The file path
     content: <data> // A Buffer, Readable Stream, Pull Stream or File with the contents of the file
+    mode: '0755' // optional string or integer mode to store the entry with. strings will be interpreted as a base 8 number
```
alanshaw: When I think about a unix mode as a string I think “rwx” “r--“ etc. Why not use this format as our stringy mode?

SPEC/FILES.md
```diff
@@ -884,6 +958,43 @@ console.log(stats)
...
+##### `ipfs.files.touch(path, [options])`
+
+Where:
+
+- `path` is the path to the file or directory to update. It might be:
+  - An existing MFS path to a file or directory (e.g. `/my-dir/a.txt`)
+  - An IPFS path (e.g. `/ipfs/QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks`)
+  - A [CID][cid] instance (e.g. `new CID('QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks')`)
+- `options` is an optional Object that might contain the following keys:
+  - `mtime` A Date object, an object with `{ sec, nsecs }` properties or the output of `process.hrtime()` or `process.hrtime.bigint()`  (default: now)
```
alanshaw:
```suggestion
  - `mtime` A ` Date` object or an object with `{ sec, nsecs }` properties or the output of `process.hrtime()` or `process.hrtime.bigint()`  (default: now)
```

SPEC/FILES.md
```diff
@@ -207,6 +219,8 @@ Returns a Pull Stream, where objects can be written of the forms
 {
   path: '/tmp/myfile.txt', // The file path
   content: <data> // A Buffer, Readable Stream, Pull Stream or File with the contents of the file
+  mode: '0755' // optional string or integer mode to store the entry with. strings will be interpreted as a base 8 number
+  mtime: <time> // A Date object, an object with `{ sec, nsecs }` properties or the output of `process.hrtime()` or `process.hrtime.bigint()
```
alanshaw:
```suggestion
  mtime: <time> // A `Date` object or an object with `{ sec, nsecs }` properties or the output of `process.hrtime()` or `process.hrtime.bigint()
```

SPEC/FILES.md
achingbrain: It's possible if you specify `dag-cbor` but the thing you'd get back probably isn't readable as you'd be expecting an instance of `DAGLink` from the `ipld-dag-pb` module and you'd get a plain JS object back.

Like this it's consistent with our other commands but it's a footgun so we should get rid of this everywhere and just make IPFS things be `dag-pb`.

SPEC/FILES.md
achingbrain: Yeah?  I have to admit I don't when setting modes, only when viewing them with `ls -l` or whatever.

In my head at least, the choice was between a literal base 10 number (e.g. `493`) or a literal base 8 number (e.g. `0755`).  Linters don't seem to like literal numbers in anything other than base 10 so that's why it's a string `'0755'` above.

It's good to be flexible though, maybe something we could add later?

SPEC/FILES.md
achingbrain: TBH we can't use `process.hrtime.bigint()` until https://github.com/ipfs/aegir/issues/487 is resolved.

```suggestion
  - `mtime` Either a ` Date` object, an object with `{ sec, nsecs }` properties or the output of `process.hrtime()`  (default: now)
```

SPEC/FILES.md
achingbrain: Same as above re: https://github.com/ipfs/aegir/issues/487

```suggestion
  - `mtime` Either a ` Date` object, an object with `{ sec, nsecs }` properties or the output of `process.hrtime()`  (default: now)
```

SPEC/FILES.md
alanshaw: If it helps to get this stuff merged faster then sure 👍 We’ll be supporting rwx format in chmod so I reckon people will expect to be able to use it here too.

SPEC/FILES.md
alanshaw: +1 with the wording but this needs formatting appropriately for the code block it’s in.

SPEC/FILES.md
alanshaw: Can you remove from the docs here then and either also remove the others or make an issue so we can track it? 🙏

achingbrain: I've pulled it out of this PR and opened #575 with removals from the other commands because `mode`/`mtime` discussions should not block that.

SPEC/FILES.md
achingbrain: We support Symbolic Notation in `chmod` (e.g. `a=rwx`, `ug-x`, etc) which are modifications to an existing `mode`.

I guess if you want to use it here you could apply the modification to the default mode for files/directories?

SPEC/FILES.md
```diff
@@ -207,6 +219,8 @@ Returns a Pull Stream, where objects can be written of the forms
 {
   path: '/tmp/myfile.txt', // The file path
   content: <data> // A Buffer, Readable Stream, Pull Stream or File with the contents of the file
+  mode: '0755' // optional string or integer mode to store the entry with. strings will be interpreted as a base 8 number
+  `mtime` Either a ` Date` object, an object with `{ sec, nsecs }` properties or the output of `process.hrtime()`  (default: now)
```
Stebalien: `mtime:`?

Also, isn't this also optional?

SPEC/FILES.md
```diff
@@ -58,6 +60,8 @@ Where `data` may be:
 {
     path: '/tmp/myfile.txt', // The file path
     content: <data> // A Buffer, Readable Stream, Pull Stream or File with the contents of the file
+    mode: '0755' // optional string or integer mode to store the entry with. strings will be interpreted as a base 8 number
+    mtime: <time> // A Date object, an object with `{ sec, nsecs }` properties or the output of `process.hrtime()` or `process.hrtime.bigint()`
```
Stebalien: optional?

SPEC/FILES.md
```diff
@@ -759,6 +789,47 @@ A great source of [examples][] can be found in the tests for this API.
...
+
+Where:
+
+- `path` is the path to the entry to modify.  It might be:
+  - An existing MFS path to a file or a directory (e.g. `/my-dir/my-file.txt`)
+  - An IPFS path (e.g. `/ipfs/QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks`)
+  - A [CID][cid] instance (e.g. `new CID('QmWGeRAEgtsHW3ec7U4qW2CyVy7eA2mFRVbk1nb24jFyks')`)
+- `mode` is the new file mode.  It might be:
+  - A string octal, e.g. `'0755'`
+  - A string modification of the existing mode, e.g. `'+x'`, `'-gw'`, etc
```
Stebalien:
```suggestion
  - A string modification of the existing mode, e.g. `'a+x'`, `'g-w'`, etc
```

SPEC/FILES.md
```diff
@@ -759,6 +789,47 @@ A great source of [examples][] can be found in the tests for this API.
...
+  - An integer, e.g. the returned value from `parseInt('0755', 8)`
```
Stebalien: Or just `0o755`? (I'm not sure if supporting the string version is really necessary given that JS has octal support but it may make for nicer JSON)?

SPEC/FILES.md
```diff
@@ -759,6 +789,47 @@ A great source of [examples][] can be found in the tests for this API.
...
+| `Promise<void>` | If action is successfully completed. Otherwise an error will be thrown |
+
+**Example:**
+
+```JavaScript
+// To give a file -rwxrwxrwx permissions
+await ipfs.files.chmod('/path/to/file.txt', parseInt('0777', 8))
+
+// Alternatively
+await ipfs.files.chmod('/path/to/file.txt', '+rwx')
```
Stebalien: `?+rwx`?

SPEC/FILES.md
achingbrain: Octals-in-strings are supported because the recommended eslint configuration [disallows octal literals](https://eslint.org/docs/rules/no-octal) - our users are likely to be passing them around as strings instead as that's what eslint deems to be the 'correct' way to do it.

SPEC/FILES.md
achingbrain: Not sure I follow - what does the leading `?` mean in `?+rwx`?  I can't see any reference to it in `man chmod` and when I try to do this locally I get an error:

```console
$ chmod ?+rwx foo.txt
chmod: Invalid file mode: ?+rwx
```

SPEC/FILES.md
achingbrain: Fixed!

SPEC/FILES.md
achingbrain: Fixed!

SPEC/FILES.md
Stebalien: It disallows octal literals of the form '07...' but I'm pretty sure '0o7...' (ES6) is perfectly fine.

SPEC/FILES.md
Stebalien: Ah, nevermind. I didn't realize `+rwx` was equivalent to `a+rwx`.

alanshaw: That's a good point, it will support `u+rwx`, `g+r`, `o+w`, `a-w` etc. though right? We should specify what subset of https://en.wikipedia.org/wiki/Modes_(Unix)#Format we do support.

achingbrain: Yes, it'll support [all of those](https://github.com/ipfs/js-ipfs-mfs/pull/63/files#diff-e68b618aab9b596a8f2ed03586f2ef2aR79-R151).

Though now you mention it there's a testing gap for `a...`

..and maybe not `X` right now, though it can be added.

SPEC/FILES.md
achingbrain: Fair point - we can drop the string octal stuff as args in that case.

At the moment the js implementation uses string octals for mode values returned from the various RPC HTTP endpoints for the sake of readability, happy to keep that?

achingbrain:
I'm going to merge this as-is since I think we're vaguely in agreement, we can address any issues in further PRs.

PR #577, opened 2020-01-07
fix: disconnect by peerid not multiaddr

jacobheun:
Disconnecting by a multiaddr isn't reliable as we're not always guaranteed to have a PeerId in the multiaddr. Using a PeerId is reliable. This updates the disconnect test to use the PeerId of the peer instead of the multiaddr

### DIFF
--- a/src/swarm/disconnect.js
+++ b/src/swarm/disconnect.js
@@ -2,6 +2,7 @@
 'use strict'
 
 const { getDescribe, getIt, expect } = require('../utils/mocha')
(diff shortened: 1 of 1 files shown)


alanshaw:
Can you rebase this onto https://github.com/ipfs/interface-js-ipfs-core/pull/567 @jacobheun as this is the PR being used in https://github.com/ipfs/js-ipfs/pull/2683

jacobheun:
@alanshaw this PR is onto that branch, do you just want me to commit to it?

PR #581, opened 2020-01-24
refactor: return peer ids as strings

achingbrain:
As part of the async iterator refactor some peer IDs were being returned as CIDs - the conversion was done in IPFS but not in libp2p, which meant that where IPFS exposes libp2p directly, peer IDs were being returned as strings, e.g. pubsub topic subscribers, message senders, etc.

The aim is to convert these to CIDs in the long run but such a change needs to be driven from inside libp2p instead of piecemeal at the IPFS layer.

This PR changes the `ipfs.id().id` and `ipfs.swarm.peers().[].peer` properties to strings so we can do the CID conversion in one go.

Also adds more in-depth tests for `ipfs.id` and documents the output from that method.

BREAKING CHANGE:

Where `PeerID`s were previously [CID]s, now they are Strings

- `ipfs.bitswap.stat().peers[n]` is now a String (was a CID)
- `ipfs.dht.findPeer().id` is now a String (was a CID)
- `ipfs.dht.findProvs()[n].id` is now a String (was a CID)
- `ipfs.dht.provide()[n].id` is now a String (was a CID)
- `ipfs.dht.put()[n].id` is now a String (was a CID)
- `ipfs.dht.query()[n].id` is now a String (was a CID)
- `ipfs.id().id` is now a String (was a CID)
- `ipfs.id().addresses[n]` are now [Multiaddr]s (were Strings)

[CID]: https://www.npmjs.com/package/cids
[Multiaddr]: https://www.npmjs.com/package/multiaddr

### DIFF
--- a/SPEC/MISCELLANEOUS.md
+++ b/SPEC/MISCELLANEOUS.md
@@ -23,7 +23,7 @@ The Peer identity has the following properties:
 
 - `id: String` - the Peer ID
 - `publicKey: String` - the public key of the peer as a base64 encoded string
-- `addresses: String[]` - A list of multiaddrs this node is listening on as strings
+- `addresses: Multiaddr[]` - A list of multiaddrs this node is listening on
 - `agentVersion: String` - The agent version
 - `protocolVersion: String` - The supported protocol version
 
--- a/SPEC/MISCELLANEOUS.md
+++ b/SPEC/MISCELLANEOUS.md
@@ -19,6 +19,14 @@
 | -------- | -------- |
 | `Promise<Object>` | An object with the Peer identity |
 
+The Peer identity has the following properties:
+
+- `id: String` - the Peer ID
+- `publicKey: String` - the public key of the peer as a base64 encoded string
+- `addresses: String[]` - A list of multiaddrs this node is listening on as strings
+- `agentVersion: String` - The agent version
+- `protocolVersion: String` - The supported protocol version
+
 **Example:**
 
 ```JavaScript
--- a/src/dht/find-peer.js
+++ b/src/dht/find-peer.js
@@ -27,14 +27,15 @@ module.exports = (common, options) => {
     after(() => common.clean())
 
     it('should find other peers', async () => {
-      const res = await nodeA.dht.findPeer(nodeB.peerId.id)
-
+      const nodeBId = await nodeB.id()
+      const res = await nodeA.dht.findPeer(nodeBId.id)
       const id = res.id.toString()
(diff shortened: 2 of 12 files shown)


src/swarm/peers.js
```diff
@@ -41,8 +40,7 @@ module.exports = (common, options) => {
 
       expect(peer).to.have.a.property('addr')
       expect(multiaddr.isMultiaddr(peer.addr)).to.equal(true)
-      expect(peer).to.have.a.property('peer')
-      expect(CID.isCID(peer.peer)).to.equal(true)
+      expect(peer).to.have.a.property('peer').that.is.a('string')
```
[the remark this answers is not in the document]
achingbrain: Have reinstated the assertion and added it to the `ipfs.id` test too.

alanshaw:
Is ready for re-review?

achingbrain:
Yes, should be good to go.

SPEC/MISCELLANEOUS.md
```diff
@@ -19,6 +19,14 @@
 | -------- | -------- |
 | `Promise<Object>` | An object with the Peer identity |
 
+The Peer identity has the following properties:
+
+- `id: String` - the Peer ID
+- `publicKey: String` - the public key of the peer as a base64 encoded string
+- `addresses: String[]` - A list of multiaddrs this node is listening on as strings
```
alanshaw:
```suggestion
- `addresses: Multiaddr[]` - A list of multiaddrs this node is listening on
```

src/dht/find-peer.js
```diff
@@ -27,14 +27,15 @@ module.exports = (common, options) => {
     after(() => common.clean())
 
     it('should find other peers', async () => {
-      const res = await nodeA.dht.findPeer(nodeB.peerId.id)
-
+      const nodeBId = await nodeB.id()
```
alanshaw: Why not use `nodeB.peerId` here?

achingbrain: It's not part of the API defined in `interface-js-ipfs-core` so we should really remove it from `ipfsd-ctl` return values.

This is my mind wandering from the task in hand though so I'm happy to change it if required.

alanshaw: Right, so it shouldn't be used in the interface tests because we can't guarantee `ipfsd-ctl` spawned the node(s) in the test. 👍 Sounds like a good plan, I'm happy to leave this like it is. Do you want to open an issue here to track that task?

achingbrain: I've opened https://github.com/ipfs/interface-js-ipfs-core/issues/582

PR #584, opened 2020-01-29
fix: do not assume certain implementations of ipfs are present

achingbrain:
Passing `type: 'js'` requires the containing project to have a dependency on js-IPFS which may not be the case.

I don't think anything in these tests requires the js implementation.  Please do shout if something does.

### DIFF
--- a/src/bitswap/wantlist.js
+++ b/src/bitswap/wantlist.js
@@ -21,7 +21,7 @@ module.exports = (common, options) => {
 
     before(async () => {
       ipfsA = (await common.spawn()).api
-      ipfsB = (await common.spawn({ type: 'go' })).api
+      ipfsB = (await common.spawn()).api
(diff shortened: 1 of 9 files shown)


alanshaw:
refs https://github.com/ipfs/interface-js-ipfs-core/issues/582#issuecomment-580215033

For `type: 'go'` it's a problem with browsre nodes not having swarm addresses to connect to. This PR needs a corresponding PR to js-ipfs to ensure that browser nodes get swarm addrs (by including webtrc transport or otherwise).

achingbrain:
> This PR needs a corresponding PR to js-ipfs to ensure that browser nodes get swarm addrs

I've added this to https://github.com/ipfs/js-ipfs/pull/2729 in https://github.com/ipfs/js-ipfs/pull/2729/commits/2b6f0e9116da773dbca6a6523a3811e1a2552c55

PR #587, opened 2020-02-04
test: adds test for adding directories with --only-hash

achingbrain:
Closes ipfs/js-ipfs#2684


alanshaw:
Thanks for picking this up and adding a test ❤️

PR #588, opened 2020-02-09
test: block put with CID string

alanshaw:
refs https://github.com/ipfs/js-ipfs/issues/2751


alanshaw:
Merging as these are passing upstream in https://github.com/ipfs/js-ipfs/pull/2760

PR #589, opened 2020-02-09
test: add meta for ipfs.get tests

alanshaw:


src/get.js
```diff
@@ -196,5 +196,27 @@ module.exports = (common, options) => {
...
+
+    it('should get file with mtime', async () => {
+      const mtime = { secs: Math.floor((Date.now() - (Math.random() * 1000)) / 1000) }
+      const content = String(Math.random() + Date.now())
+      const [{ cid }] = await all(ipfs.add({ content, mtime }))
+
+      const res = await all(ipfs.get(cid))
+
+      expect(res).to.have.length(1)
+      expect(res).to.have.nested.property('0.mtime.secs').that.deep.equals(mtime.secs)
```
achingbrain: Can you add an extra test where the `mtime` passed to `ipfs.add` contains a fractional nanosecond and asserts that `nsecs` is generated properly.

